PR #1, opened 2017-05-24
NanoHTTPD server + basic espresso call

dpgraham:


android-server-app/app/src/androidTest/java/io/appium/espressoserver/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,94 @@
+package io.appium.espressoserver;
```
mykola-mokhnach: it might be a good idea to add license header to each .java file

android-server-app/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="io.appium.espressoserver">
+
+    <application android:allowBackup="true" android:icon="@mipmap/ic_launcher"
```
jlipps: do we actually have to include icons?

android-server-app/app/src/test/java/io/appium/espressoserver/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
+package io.appium.espressoserver;
```
jlipps: do we need this file? is this in anticipation of writing unit tests?

android-server-app/build.gradle
```diff
@@ -0,0 +1,26 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
+
+    repositories {
+        maven { url 'https://maven.google.com' }
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.0.0-alpha1'
```
jlipps: should we be using an alpha version of gradle?

android-server-app/app/src/androidTest/java/io/appium/espressoserver/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Instrumentation test, which will execute on an Android device.
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+@RunWith(AndroidJUnit4.class)
+@LargeTest
+public class ExampleInstrumentedTest {
+
+    public class AppiumResponse {
```
jlipps: why are we putting classes in classes?

android-server-app/app/src/androidTest/java/io/appium/espressoserver/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,94 @@
...
+     * <p>
+     * Rules are interceptors which are executed for each test method and will run before
+     * any of your setup code in the {@link Before @Before} method.
+     * <p>
+     * {@link ActivityTestRule} will create and launch of the activity for you and also expose
+     * the activity under test. To get a reference to the activity you can use
+     * the {@link ActivityTestRule#getActivity()} method.
+     */
+    @Rule
+    public ActivityTestRule<MainActivity> mActivityRule = new ActivityTestRule<>(
```
jlipps: this will be deleted and updated with @imurchie's strategy but fine for now

android-server-app/app/src/androidTest/java/io/appium/espressoserver/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,94 @@
...
+            MainActivity.class);
+
+    @Test
+    public void changeText_newActivity() throws InterruptedException, IOException {
```
jlipps: should probably call this something else more descriptive like `startEspressoServer`

android-server-app/app/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
dpgraham: Yeah, we can move it over. This was built by default.

android-server-app/app/src/androidTest/java/io/appium/espressoserver/ExampleInstrumentedTest.java
dpgraham: It's just a PoC for now. I'll be moving everything out afterwards.

imurchie:
Or just `server`, since it is clear this is Espresso, given the driver name.

android-server-app/app/src/test/java/io/appium/espressoserver/ExampleUnitTest.java
dpgraham: This is generated by default. I think we should leave it there in the event that we want to write  unit tests and this is there as a stub.

android-server-app/app/src/main/AndroidManifest.xml
dpgraham: No I'll get rid of them.

jlipps:
> Or just server, since it is clear this is Espresso, given the driver name.

Yes this works too. However I wouldn't want people to get confused since there will be two servers in this project: the node+appium proxy server, then the server running in the espresso context. Just `server` on its own is clear to me, but I'm wondering if it might not be clear initially to others.

imurchie:
> but I'm wondering if it might not be clear initially to others

Reasonable enough. I'm down with `espresso-server`.

dpgraham:
I'm just assuming we need to commit the gradle-wrapper jar because it was part of the original project and wasn't included in the default .gitignore

PR #4, opened 2017-05-24
Added router to server

dpgraham:
* Moved Espresso into normal dependencies instead of test dependencies
* Added models, http, router, exceptions to main project
* Wrote a simple router that maps URI + Method to a handler
* Wrote one handler that takes a ‘selector’ queryString and then looks for an element that has a matching content-desc

### DIFF
--- a/espresso-server/app/app.iml
+++ b/espresso-server/app/app.iml
@@ -99,32 +99,32 @@
     <orderEntry type="jdk" jdkName="Android API 25 Platform" jdkType="Android SDK" />
     <orderEntry type="sourceFolder" forTests="false" />
     <orderEntry type="library" exported="" name="gson-2.2.4" level="project" />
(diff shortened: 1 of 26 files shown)


imurchie:
What was behind the decision to move all this into the App instead of the Test? I personally would opt for the App being a skeleton that does nothing (indeed, we shouldn't even need to install it), and everything happening in the test.

dpgraham:
Because I want to be able to write tests for it and it would be confusing if we had tests that imported test code. Also I don't think that we could do code coverage analysis.

Technically, I don't think it makes a difference being inside the app folder. The code that is in `app` doesn't actually get imported by MainActivity, it only gets imported by the test file.

imurchie:
I'm not sold.

Why not have it in a library, then, instead of in the app that doesn't use it?

dpgraham:
Changes made

PR #6, opened 2017-05-25
Only rebuild server if necessary

imurchie:
If we already have an espresso server apk with the correct version and instrumented for the correct package, we don't need to build it again.

But we want to be able to test changes we've made in development, so add the internal cap `forceEspressoRebuild` to allow us to blow away the one that is already there.

Also add the pre-commit hooks and fix some linting issues.

### DIFF
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -32,13 +32,10 @@ class EspressoRunner {
       await fs.unlink(this.modServerPath);
     }
 
-    let needsUninstall = false;
     if (!(await fs.exists(this.modServerPath))) {
       await this.buildNewModServer();
-      needsUninstall = true;
     }
-    needsUninstall = await this.checkAndSignCert(this.modServerPath) || needsUninstall;
-    if (needsUninstall) {
+    if (await this.checkAndSignCert(this.modServerPath)) {
       logger.info("New server was built, uninstalling any instances of it");
       await this.adb.uninstallApk(TEST_APK_PKG);
     }
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -3,18 +3,19 @@ import { JWProxy } from 'appium-base-driver';
 import { retryInterval } from 'asyncbox';
 import logger from './logger';
 import path from 'path';
-import { fs } from 'appium-support';
+import { fs, util } from 'appium-support';
+import { version } from '../../package.json'; // eslint-disable-line import/no-unresolved
 
 
 const TEST_APK_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
 const TEST_MANIFEST_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'AndroidManifest-test.xml');
 const TEST_APK_PKG = "io.appium.espressoserver.test";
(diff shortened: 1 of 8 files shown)


lib/espresso-runner.js
```diff
@@ -3,38 +3,47 @@ import { JWProxy } from 'appium-base-driver';
...
     }
 
-    let tmpTestApkPath = await this.buildNewModServer();
-    await this.signAndInstall(this.modServerPath, TEST_APK_PKG);
+    let needsUninstall = false;
+    if (!(await fs.exists(this.modServerPath))) {
+      await this.buildNewModServer();
+      needsUninstall = true;
+    }
+    needsUninstall = await this.checkAndSignCert(this.modServerPath) || needsUninstall;
```
jlipps: if we do this the other way around `needsUninstall = needsUninstall || ...` we would be able to take advantage of short-circuiting and not run the cert check if we don't have to.

lib/espresso-runner.js
```diff
@@ -3,38 +3,47 @@ import { JWProxy } from 'appium-base-driver';
...
-      await this.adb.uninstallApk(TEST_APK_PKG);
+    if (this.forceEspressoRebuild) {
+      logger.debug(`Capability 'forceEspressoRebuild' on. Deleting file '${this.modServerPath}'`);
+      await fs.unlink(this.modServerPath);
     }
 
-    let tmpTestApkPath = await this.buildNewModServer();
-    await this.signAndInstall(this.modServerPath, TEST_APK_PKG);
+    let needsUninstall = false;
+    if (!(await fs.exists(this.modServerPath))) {
```
mykola-mokhnach: probably, we need to be a bit smarter there and rebuild the apk if it has older version (for example after the driver upgrade happened)

lib/espresso-runner.js
imurchie: But the check actually signs. So if we have built a new server apk, we will need to do the signing, even though `needsUninstall` is true. I think the actual fact is that the `|| needsUninstall` is superfluous... if we built a new server `checkAndSignCert` will _always_ return `true`. So we could get rid of `needsUninstall` and just have the cert check inside the following conditional.

lib/espresso-runner.js
imurchie: The driver version is encoded in `this.modServerPath`.

lib/espresso-runner.js
jlipps: ah yes i forgot that side effect was essential in any case. your new logic sounds good tome.

lib/espresso-runner.js
imurchie: Extract what? 

The value of `this.modServerPath` is of the form `/path/to/io.appium.espressoserver.test_<version of driver>_<package under test>.apk`.

So for our ApiDemos, it is `/path/to/io.appium.espressoserver.test_0.0.1_io.appium.android.apis.apk`. When we upgrade, say to `appium-espresso-driver` `0.0.2`, it will become `/path/to/io.appium.espressoserver.test_0.0.2_io.appium.android.apis.apk`, which will not be found, and trigger a rebuild with the latest server code.

mykola-mokhnach: ok, got it. I thought under "encoded" you mean "is stored in apk manifest"

imurchie: Oh! No. See line 26 above.

PR #15, opened 2017-05-26
Make the Appium model the base response for all responses

dpgraham:
* Appium has `status`, `sessionId`, `id` and `value` as it’s parameters
* Every object should extend this model


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Finder.java
```diff
@@ -13,6 +13,7 @@
 import io.appium.espressoserver.lib.Exceptions.InvalidStrategyException;
 import io.appium.espressoserver.lib.Exceptions.ServerErrorException;
 import io.appium.espressoserver.lib.Http.Response.AppiumResponse;
+import io.appium.espressoserver.lib.Model.Appium;
```
jlipps: is `Appium` a better name than `AppiumResponse`?

dpgraham: Appium is the serializable model and AppiumResponse is the HTTP response.

What I could do is rename Appium to AppiumResponse and then just change AppiumResponse to BaseResponse since _all_ responses have the same format.

PR #18, opened 2017-05-26
Added appium statuses; made lint fixes

dpgraham:
* Added all of the different appium status error types
* Also made fixes for problems found by linting tool


dpgraham:
@mykola-mokhnach That's a good idea, I'll make an issue for that

PR #21, opened 2017-05-26
Initial e2e test

imurchie:
Begin on an initial e2e test that finds a text field and types into it.

There is a problem with the parameter parsing. The following error is returned:
```
info HTTP --> POST /wd/hub/session/f982a4a3-dcba-4f89-9ae8-140626c4c93f/element {"using":"id","value":"edit1"}
info MJSONWP Driver proxy active, passing request on via HTTP proxy
dbug JSONWP Proxy Proxying [POST /wd/hub/session/f982a4a3-dcba-4f89-9ae8-140626c4c93f/element] to [POST http://localhost:8080/session/3d20191f-b2a9-4db6-8817-e7c8c1a99877/element] with body: {"using":"id","value":"edit1"}
ERR! MJSONWP Encountered internal error running command: Error: Could not proxy. Proxy error: Could not proxy command to remote server. Original error: 500 - {"value":{"reason":"Attempt to invoke interface method 'java.lang.Object java.util.List.get(int)' on a null object reference","value":{}}}
ERR! MJSONWP     at doJwpProxy$ (../../../lib/mjsonwp/mjsonwp.js:354:13)
```

And when inspected the parameters are empty here https://github.com/appium/appium-espresso-driver/blob/master/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Finder.java#L55


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -51,6 +51,8 @@ public BaseResponse route(IHTTPSession session) {
             String uri = session.getUri();
             Method method = session.getMethod();
 
+            System.out.println("Received " + method + " request for '" + uri + "'");
+
```
dpgraham: We should replace these with proper logs at some point

PR #23, opened 2017-05-26
Some warts, but a running e2e test that does stuff

imurchie:
This introduces a lot of redundancy around parameters handling that will have to be refactored, and there is the weirdness with NanoHTTPD sometimes failing to understand the method.

But it's a functional test that finds an element, clicks it, and types.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/SendKeys.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/SendKeys.java
@@ -3,6 +3,8 @@ package io.appium.espressoserver.lib.Handlers;
 import android.support.test.espresso.PerformException;
 import android.support.test.espresso.ViewInteraction;
 
+import com.google.gson.Gson;
+
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/SendKeys.java
```diff
@@ -35,16 +38,20 @@ public BaseResponse handle(NanoHTTPD.IHTTPSession session, Map<String, String> u
 
         // NanoHTTP requires call to parse body before we can get the parameters
         // TODO: Move parameter parsing into Router.java
+        Map<String, List<String>> parameters;
         try {
-            session.parseBody(new HashMap<String, String>());
+            Map<String, String> files = new HashMap<String, String>();
```
mykola-mokhnach: there's no need to set template types twice:

```
Map<String, String> files = new HashMap<>();
```

test/functional/commands/keyboard-e2e-specs.js
```diff
@@ -22,9 +22,15 @@ describe('createSession', function () {
   });
 
   it('should start android session focusing on default activity', async () => {
-    let el = await driver.elementById('edit1');
+    let el = await driver.elementById('id/edit1');
```
mykola-mokhnach: shouldn't "id/" prefix be added automatically?

imurchie: Yes this was, again, just a POC and we wanted to get an actual functional test in. Once we come back to this and start work on it, if indeed we ever do, these things can be addressed.

PR #25, opened 2017-05-27
Remove 20s delay when creating instrumentation process

imurchie:
Stop `await`ing the `exec` of the instrumentation process. This makes error handling a bit wonky, but has the advantage of stripping 20s off each session.

### DIFF
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -1,4 +1,3 @@
-import { exec } from 'teen_process';
 import { JWProxy } from 'appium-base-driver';
 import { retryInterval } from 'asyncbox';
 import logger from './logger';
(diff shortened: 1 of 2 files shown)


lib/espresso-runner.js
```diff
@@ -65,46 +64,30 @@ class EspressoRunner {
...
+    });
 
     logger.info('Waiting for Espresso to be online...');
+
     // wait 20s for espresso to be online
     await retryInterval(20, 1000, async () => {
+      if (err) return; // eslint-disable-line curly
       await this.jwproxy.command('/status', 'GET');
     });
+    if (err) throw err; // eslint-disable-line curly
```
mykola-mokhnach: it's probably clear here, that instruments have been successfully started, so it should be safer to await it here in comparison to just checking the result of err:

```
let err;
let instrumentsExecutor =  this.adb.shell(cmd).catch((e) => { // eslint-disable-line
      // handle asynchronous errors that no longer get thrown
      err = e;
    });
...
await instrumentsExecutor;
```

PR #28, opened 2017-05-30
Refactor handlers

dpgraham:
* Changed handlers so that they don't deal with HTTPSessions or HTTP responses, it only returns objects or throws exceptions and then the router returns the appropriate response. Decouples Handler from having to deal with HTTP.
* Added AppiumException and exceptions that extend it. Handlers throw these exceptions, the Router catches them and then responds with the appropriate HTTP Status code
* Instead of using Maps for params, params are defined as a class that extend AppiumParams. GSON serializes the JSON body to those classes. Less risk of runtime error this way (no hardcoded parameter names) and better readability.
* Move Session verification into the Router before the Handler is called. If no valid session, throws an exception.


imurchie:
This seems good (though I didn't go through with too fine toothed a comb.

Is it possible to either rename packages or change Codacy's intense desire to have that done? If it just complains on every PR it is not of much use.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/RequestHandler.java
```diff
@@ -1,10 +1,8 @@
...
+import io.appium.espressoserver.lib.Handlers.Exceptions.AppiumException;
+import io.appium.espressoserver.lib.Model.AppiumParams;
 
-import fi.iki.elonen.NanoHTTPD.IHTTPSession;
-import io.appium.espressoserver.lib.Http.Response.BaseResponse;
-
-public interface RequestHandler {
-    BaseResponse handle(IHTTPSession session, Map<String, String> uriParams);
+public interface RequestHandler<T extends AppiumParams>{
+    Object handle(T params) throws AppiumException;
```
mykola-mokhnach: Why Object? Cannot the retuned result be generic as well?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Exceptions/AppiumException.java
```diff
@@ -0,0 +1,12 @@
+package io.appium.espressoserver.lib.Handlers.Exceptions;
+
+public abstract class AppiumException extends Exception {
```
mykola-mokhnach: Is there any reason to make it abstract?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Click.java
```diff
@@ -2,47 +2,22 @@
...
-                return new BadRequestResponse("Could not find element with ID: " + id);
-            }
-        } else {
-            return new BadRequestResponse("Could not find element with ID: " + id);
+    public Object handle(AppiumParams params) throws AppiumException {
+        ViewInteraction viewInteraction = Element.getById(params.getElementId());
+        try {
+            viewInteraction.perform(click());
+        } catch (Exception e) { // TODO: Can we narrow down these exceptions?
+            return new AppiumException("Could not find element " + params.getElementId()) {
```
mykola-mokhnach: I don't get it - do we really want to create an instance of abstract class?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Exceptions/InvalidStrategyException.java
```diff
@@ -0,0 +1,8 @@
+package io.appium.espressoserver.lib.Handlers.Exceptions;
+
+
```
mykola-mokhnach: extra empty line

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Finder.java
```diff
@@ -31,95 +21,54 @@
...
-        } catch (final InvalidStrategyException e) {
-            response.setResponse(new Appium());
-            return response;
-        }
-
-        // Get the description
-        String selector = parameters.get("value").get(0);
+public class Finder implements RequestHandler<Locator> {
 
+    public Element handle(Locator locator) throws AppiumException {
```
mykola-mokhnach: missing \@Override annotation. Same commend to other overriden methods

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Click.java
dpgraham: This was a mistake.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Finder.java
```diff
@@ -31,95 +21,54 @@
...
-            return response;
-        } catch (InvalidStrategyException e) {
-            return new BadRequestResponse(e.getMessage());
-        } catch (ServerErrorException e) {
-            return new InternalErrorResponse(e.getMessage());
-        } catch (ElementNotFoundException e) {
-            return new BadRequestResponse("Could not find element with " + strategy.getStrategyName() + ": " + selector);
+            return new Element(matcher);
+        } catch (NoMatchingViewException e) {
+            throw new NoSuchElementException("Could not find element with strategy " + locator.getUsing() + " and selector " + locator.getValue());
```
mykola-mokhnach: I'd rather use String.format

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/SendKeys.java
```diff
@@ -3,71 +3,34 @@
...
-            Gson gson = new Gson();
-            parameters = gson.fromJson(files.get("postData"), Map.class);
-        } catch (NanoHTTPD.ResponseException e) {
-            return new BadRequestResponse("Could not parse parameters");
-        } catch (IOException e) {
-            return new InternalErrorResponse("Internal server error has occurred");
+        // Convert the array of text to a String
+        String[] textArray = params.getValue();
+        StringBuilder stringBuilder = new StringBuilder();
+        for (String text: textArray) {
```
mykola-mokhnach: One can also use StringUtils.join from http://commons.apache.org/proper/commons-lang/ for this purpose

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Response/AppiumResponse.java
```diff
@@ -1,27 +1,62 @@
...
+
+    AppiumResponse(AppiumStatus status, Object value) {
+        init(status, value, null);
+    }
+
+    public AppiumResponse(AppiumStatus status, Object value, String sessionId) {
+        init(status, value, sessionId);
+    }
+
+    private void init(AppiumStatus status, Object value, String sessionId) {
```
mykola-mokhnach: one can simply [call one constructor from another](https://stackoverflow.com/questions/285177/how-do-i-call-one-constructor-from-another-in-java). No need to use a separate method for this purpose.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Response/AppiumResponse.java
```diff
@@ -1,27 +1,62 @@
...
+        this.status = status;
+        this.sessionId = sessionId;
+        id = UUID.randomUUID().toString();
+
+        switch (status) {
+            case SUCCESS:
+                httpStatus = Status.OK;
+                break;
+            case UNKNOWN_COMMAND:
+                httpStatus = Status.NOT_FOUND;
```
mykola-mokhnach: missing break?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Response/AppiumResponse.java
```diff
@@ -1,27 +1,62 @@
...
-public class AppiumResponse extends BaseResponse {
-    public AppiumResponse() {
-        setResponse(new Appium());
+@SuppressWarnings("unused")
+public class AppiumResponse {
+    private transient Status httpStatus;
+    private Object value;
+    private AppiumStatus status;
+    private String sessionId;
+    private String id; // Unique Appium transaction ID
```
mykola-mokhnach: lint does not like such comments style. Usually comments are put right before the line they are related to

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Response/AppiumResponse.java
```diff
@@ -1,27 +1,62 @@
...
+import fi.iki.elonen.NanoHTTPD.Response.Status;
 import io.appium.espressoserver.lib.Model.AppiumStatus;
 
-public class AppiumResponse extends BaseResponse {
-    public AppiumResponse() {
-        setResponse(new Appium());
+@SuppressWarnings("unused")
+public class AppiumResponse {
+    private transient Status httpStatus;
+    private Object value;
```
mykola-mokhnach: Can **value** be generic?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -1,113 +1,192 @@
...
+import io.appium.espressoserver.lib.Model.AppiumStatus;
+import io.appium.espressoserver.lib.Model.Element;
+import io.appium.espressoserver.lib.Model.Locator;
+import io.appium.espressoserver.lib.Model.Session;
+import io.appium.espressoserver.lib.Model.SessionParams;
+import io.appium.espressoserver.lib.Model.TextParams;
 
 class Router {
     private final Map<Method, HashMap<String, RequestHandler>> routerMap;
+    private final Map<Method, HashMap<String, Class>> paramClassMap;
```
mykola-mokhnach:
```java
private final Map<Method, Map<String, RequestHandler>> routerMap = new ConcurrentHashMap<>();
private final Map<Method, Map<String, ?>> paramClassMap = new ConcurrentHashMap<>();
```

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -1,113 +1,192 @@
...
+            String testUri = entry.getKey();
+
+            // TODO: Use StringBuilder to construct the Test Regexes
+            String testRegex = "^";
+            Map<Integer, String> wildcardIndices = new HashMap<>();
+
+            // Convert route to a regex to test incoming URI against
+            // TODO: Cache these regexes instead of re-creating them every time
+            int index = 0;
+            for (String uriToken : testUri.split("/")) {
```
mykola-mokhnach: this method is a bit hard to read. Can we extract parsing operations into separate private helper methods?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -1,113 +1,192 @@
...
+
+        // Parse it to an Appium param
+        Gson gson = new Gson();
+        String postJson = parseBody(session);
+        AppiumParams appiumParams;
+        // TODO: Need to find a way to specify 'required' fields and throw exception when not provided
+        if (postJson == null) {
+            appiumParams = new AppiumParams();
+        } else {
+            appiumParams = (AppiumParams) paramClass.cast(gson.fromJson(postJson, paramClass));
```
mykola-mokhnach: 👍

mykola-mokhnach: We can create Gson instance here, because it is not used anywhere else (or, probably, cache it on class level)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/EspressoServerRunnerTest.java
```diff
@@ -21,9 +21,9 @@
 public class EspressoServerRunnerTest {
 
     @Test
-    public void startEspressoServer() throws InterruptedException, IOException, ServerErrorException {
+    public void startEspressoServer() throws InterruptedException, IOException, DuplicateRouteException {
         new Server();
         // TODO: Figure out how to keep Runner open forever
-        Thread.sleep(300000);
+        Thread.sleep(3000000);
```
mykola-mokhnach: https://www.reddit.com/r/androiddev/comments/2r3g8k/keep_android_awake_for_incoming_network/

```java
@Override
public int onStartCommand(Intent intent, int flags, int startId) {

    StartHttpServer();
    ShowNotification();
    return Service.START_STICKY;
}
```

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Model/Element.java
```diff
@@ -20,7 +20,11 @@ public String getElementId() {
         return ELEMENT;
     }
 
-    public static Map<String, ViewInteraction> getCache() {
-        return cache;
+    public static ViewInteraction getById(String elementId) {
+        return cache.get(elementId);
```
mykola-mokhnach: Probably, cache itself should be thread-safe as well?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Exceptions/AppiumException.java
dpgraham: Not really. I just figured it's not going to ever be used directly so declare it abstract. I could make it an interface, I suppose.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Response/AppiumResponse.java
dpgraham: It is, yes. I'm surprised that lint doesn't catch that.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -33,12 +34,12 @@
 import io.appium.espressoserver.lib.Model.TextParams;
 
 class Router {
-    private final Map<Method, HashMap<String, RequestHandler>> routerMap;
-    private final Map<Method, HashMap<String, Class>> paramClassMap;
+    private final Map<Method, Map<String, RequestHandler>> routerMap;
+    private final Map<Method, Map<String, Class>> paramClassMap;
```
mykola-mokhnach: Class<? extends AppiumParams>

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -76,7 +77,7 @@ AppiumResponse route(IHTTPSession session) {
         System.out.println("Received " + method + " request for '" + uri + "'");
 
         if (!routerMap.containsKey(method)) {
-            routerMap.put(method, new HashMap<String, RequestHandler>());
+            routerMap.put(method, new ConcurrentHashMap<String, RequestHandler>());
         }
         if (!paramClassMap.containsKey(method)) {
             paramClassMap.put(method, new HashMap<String, Class>());
```
mykola-mokhnach: new HashMap<>()

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -76,7 +77,7 @@ AppiumResponse route(IHTTPSession session) {
         System.out.println("Received " + method + " request for '" + uri + "'");
 
         if (!routerMap.containsKey(method)) {
-            routerMap.put(method, new HashMap<String, RequestHandler>());
+            routerMap.put(method, new ConcurrentHashMap<String, RequestHandler>());
```
mykola-mokhnach: there is no need to define generic types for the constructor again

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/RequestHandler.java
```diff
@@ -1,10 +1,8 @@
...
-import java.util.Map;
+import io.appium.espressoserver.lib.Handlers.Exceptions.AppiumException;
+import io.appium.espressoserver.lib.Model.AppiumParams;
 
-import fi.iki.elonen.NanoHTTPD.IHTTPSession;
-import io.appium.espressoserver.lib.Http.Response.BaseResponse;
-
-public interface RequestHandler {
-    BaseResponse handle(IHTTPSession session, Map<String, Object> params);
+public interface RequestHandler<T extends AppiumParams>{
```
mykola-mokhnach:
```java
public interface RequestHandler<T extends AppiumParams, R extends ...> {
  R handle(T params) throws AppiumException;
}
```

General rule here - always try to "generalise" basic common types like Class and Object. Let compiler do the dirty job for you

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
dpgraham: I think in this case the generics are required because we're putting a HashMap into a Map as a value. When I took the generics out it caused an error.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/RequestHandler.java
dpgraham: Good call.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Finder.java
dpgraham: Yeah, I'll do a separate commit for that. There are multiple places where String concatenation is used where String.Format should be used instead.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/EspressoServerRunnerTest.java
dpgraham: I made a separate issue for that. I don't think this will work because it's a Test APK that's being generated and there's no Activity or Context.

dpgraham:
Changes made

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Finder.java
jlipps: yeah, let's make an issue to fix those and add a lint rule that objects to concatenation if possible

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/SendKeys.java
```diff
@@ -3,43 +3,35 @@
...
         }
 
-        response.setAppiumStatus(AppiumStatus.SUCCESS);
+        String textValue = stringBuilder.toString();
 
-        return response;
+        try {
+            viewInteraction.perform(typeText(textValue));
+        } catch (PerformException e) {
+            return new BadParametersException("Could not apply sendKeys to element " + id + ": " + e.getMessage());
```
jlipps: why is this a BadParametersException? Wouldn't it be a generic exception?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Response/AppiumResponse.java
```diff
@@ -1,27 +1,64 @@
...
+            case SUCCESS:
+                httpStatus = Status.OK;
+                break;
+            case UNKNOWN_COMMAND:
+                httpStatus = Status.NOT_FOUND;
+                break;
+            default:
+                httpStatus = Status.BAD_REQUEST;
+                break;
+        }
```
jlipps: don't we need a generic 500 error status?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -98,32 +119,78 @@ public BaseResponse handle(IHTTPSession session, Map<String, Object> params) {
...
+        appiumParams.setElementId(uriParams.get("elementId"));
+
+        // Validate the sessionId
+        if (appiumParams.getSessionId() != null && !appiumParams.getSessionId().equals(Session.getGlobalSessionId())) {
+            return new ErrorResponse(AppiumStatus.BAD_PARAMETERS_ERROR, "Invalid session ID " + appiumParams.getSessionId());
+        }
+
+        // Validate the elementId
+        // TODO: Add a method to check if element is stale
+        if (appiumParams.getElementId() != null && !Element.exists(appiumParams.getElementId())) {
```
jlipps: we might also want to validate that the handler being called is one that cares about elements to begin with. just because someone passes in an element id doesn't mean we're actually at a handler that can do anything to an element

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Model/AppiumStatus.java
```diff
@@ -5,32 +5,34 @@
...
+    SCRIPT_TIMEOUT(28, "A script did not complete before its timeout expired"),
+    INVALID_ELEMENT_COORDINATES(29, "The coordinates provided to an interactions operation are invalid"),
+    IME_NOT_AVAILABLE(30, "IME was not available"),
+    IME_ENGINE_ACTIVATION_FAILED(31, "An IME engine could not be started"),
+    INVALID_SELECTOR(32, "Argument was an invalid selector (error.g. XPath/CSS)"),
+    SESSION_NOT_CREATED_EXCEPTION(33, "A new session could not be created"),
+    MOVE_TARGET_OUT_OF_BOUNDS(34, "Target provided for a move action is out of bounds"),
+    JSON_DECODER_ERROR(35, "Could not decode action/params of command, please check format!"),
+    INVALID_CONTEXT_ERROR(36, "The operation could not be performed in the current context"),
+    BAD_PARAMETERS_ERROR(37, "The parameters specified for the operation are incorrect");
```
jlipps: is this an actual JSONWP status? i don't remember seeing it before

jlipps: (these numbers are defined by the JSONWP spec so we can't just make up new ones)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
dpgraham: In that case, we should have the handlers extend some abstract 'ElementHandler' object which does the checking.

dpgraham: Better yet, I'll refactor the Element class to throw an exception if you attempt to get an element that doesn't exist.

jlipps: yes, good idea!

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Click.java
```diff
@@ -2,37 +2,22 @@
...
-import io.appium.espressoserver.lib.Http.Response.BaseResponse;
-import io.appium.espressoserver.lib.Model.AppiumStatus;
+import io.appium.espressoserver.lib.Handlers.Exceptions.AppiumException;
+import io.appium.espressoserver.lib.Model.AppiumParams;
 import io.appium.espressoserver.lib.Model.Element;
 
 import static android.support.test.espresso.action.ViewActions.click;
 
-public class Click extends BaseHandler {
+public class Click implements RequestHandler<AppiumParams, Object> {
```
mykola-mokhnach: replace Object with Void

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/DeleteSession.java
```diff
@@ -1,21 +1,13 @@
...
-import io.appium.espressoserver.lib.Http.Response.BaseResponse;
-import io.appium.espressoserver.lib.Model.AppiumStatus;
+import io.appium.espressoserver.lib.Model.AppiumParams;
 import io.appium.espressoserver.lib.Model.Session;
 
-import fi.iki.elonen.NanoHTTPD;
-import io.appium.espressoserver.lib.Http.Response.AppiumResponse;
-
-public class DeleteSession implements RequestHandler {
+public class DeleteSession implements RequestHandler<AppiumParams, Object> {
```
mykola-mokhnach: Object->Void

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/ScrollTo.java
```diff
@@ -1,38 +1,30 @@
...
+import io.appium.espressoserver.lib.Handlers.Exceptions.AppiumException;
+import io.appium.espressoserver.lib.Handlers.Exceptions.NoSuchElementException;
+import io.appium.espressoserver.lib.Model.ScrollToParams;
 
 import static android.support.test.espresso.Espresso.onView;
 import static android.support.test.espresso.action.ViewActions.scrollTo;
 import static android.support.test.espresso.matcher.ViewMatchers.withText;
 
-public class ScrollTo extends BaseHandler {
+public class ScrollTo implements RequestHandler<ScrollToParams, Object> {
```
mykola-mokhnach: Object->Void

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/SendKeys.java
```diff
@@ -3,43 +3,36 @@
...
+import io.appium.espressoserver.lib.Handlers.Exceptions.AppiumException;
+import io.appium.espressoserver.lib.Handlers.Exceptions.BadParametersException;
 import io.appium.espressoserver.lib.Model.Element;
+import io.appium.espressoserver.lib.Model.TextParams;
 
 import static android.support.test.espresso.action.ViewActions.typeText;
 
 
-public class SendKeys extends BaseHandler {
+public class SendKeys implements RequestHandler<TextParams, Object> {
```
mykola-mokhnach: Object->Void

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Handlers/Status.java
```diff
@@ -1,19 +1,11 @@
...
-import java.util.Map;
-
-import fi.iki.elonen.NanoHTTPD;
-
-import io.appium.espressoserver.lib.Http.Response.AppiumResponse;
-import io.appium.espressoserver.lib.Http.Response.BaseResponse;
-import io.appium.espressoserver.lib.Model.AppiumStatus;
+import io.appium.espressoserver.lib.Model.AppiumParams;
 
 public class Status implements RequestHandler {
```
mykola-mokhnach: generic type is missing

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -8,47 +8,72 @@
...
+     * Registers a route to a handler
+     * @param method HTTP method type
+     * @param uri URI of endpoint
+     * @param handler RequestHandler object that takes params and returns a result based on params
+     * @param paramClass Parameters class that the JSON is deserialized to
+     * @throws DuplicateRouteException If the same route is registered twice, throw an error
+     */
+    private void addRoute(Method method, String uri, RequestHandler handler, Class<? extends AppiumParams> paramClass) throws DuplicateRouteException {
         if (!routerMap.containsKey(method)) {
             routerMap.put(method, new ConcurrentHashMap<String, RequestHandler>());
```
mykola-mokhnach: generic type is missing for RequestHandler

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -58,19 +83,15 @@ BaseResponse route(IHTTPSession session) {
...
-                return new NotFoundResponse();
-            }
-        };
-
-        Map<String, Object> params = parseBody(session);
+        Map<String, String> uriParams = new HashMap<>();
 
         // Look for a matching route
         // TODO: Move this to a separate method 'isRouteMatch'.
+        Class paramClass = AppiumParams.class;
```
mykola-mokhnach: It is recommended to always make Class parametrised:

```
Class<? extends AppiumParams> paramClass
```

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
```diff
@@ -58,19 +83,15 @@ BaseResponse route(IHTTPSession session) {
...
         for (Map.Entry<String, RequestHandler> entry : routerMap.get(method).entrySet()) {
             String testUri = entry.getKey();
```
mykola-mokhnach: no chance to extract parsing logic to a separate method?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Server.java
```diff
@@ -3,35 +3,41 @@
...
+        GsonBuilder gsonBuilder = new GsonBuilder().serializeNulls();
+        try {
+            AppiumResponse response = router.route(session);
         gsonBuilder.registerTypeAdapter(AppiumStatus.class, new AppiumStatusAdapter());
-        // TODO: Don't hardcode application/json change it to MediaType http://docs.oracle.com/javaee/6/api/javax/ws/rs/core/MediaType.html
-        return newFixedLengthResponse(response.getHttpStatus(),  "application/json", gsonBuilder.create().toJson(response.getResponse()));
+            // TODO: Don't hardcode application/json change it to MediaType http://docs.oracle.com/javaee/6/api/javax/ws/rs/core/MediaType.html
+            return newFixedLengthResponse(response.getHttpStatus(), "application/json", gsonBuilder.create().toJson(response));
+        } catch (Exception e) {
+            StringWriter sw = new StringWriter();
```
mykola-mokhnach: maybe it worth to extract it into a separate utility method. Might be useful elsewhere.
Also, you may use https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/exception/ExceptionUtils.html#getStackTrace(java.lang.Throwable) for the same purpose

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Router.java
dpgraham: I have some TODOs there. This class will need some cleanup in the future.

dpgraham:
Revisions pushed.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/Http/Response/BaseResponse.java
```diff
@@ -1,32 +1,12 @@
...
 
-import fi.iki.elonen.NanoHTTPD;
-import io.appium.espressoserver.lib.Model.Appium;
 
-public abstract class BaseResponse {
+import fi.iki.elonen.NanoHTTPD.Response.Status;
 
-    private NanoHTTPD.Response.Status httpStatus;
-    private Appium response;
+public class BaseResponse {
```
mykola-mokhnach: why not abstract anymore?

PR #33, opened 2017-05-31
Codacy fixes

dpgraham:
* Move field declarations to top of class
* Remove one underscore
* Make on field final


dpgraham:
@mykola-mokhnach I like it so far. The only annoying thing is that Codacy and the Android Studio linter have different opinions on using package-private. But I could probably just update the lint.xml to suppress that error.

How do you like it yourself?

mykola-mokhnach:
Yes, it does what it supposed to do. However the PR verification performance is quite low, especially for big PRs. Probably, it's because open-source (free) plan is used.

PR #35, opened 2017-05-31
Edit Session class

dpgraham:
* Made it so that you can’t instantiate Session, you can only call `createGlobalSession`, `deleteGlobalSession` and `getGlobalSessionID` from static context
* `createGlobalSession` returns an instance of Session with just an `id` field. This object is needed for GSON serialization.
* If `createGlobalSession` is called while a session is in progress, throws an exception. We can only run one at a time.
* This fixes a Codacy complaint of static variable being accessed inside of a constructor.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Session.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Session.java
@@ -8,7 +8,7 @@ import io.appium.espressoserver.lib.handlers.exceptions.SessionNotCreatedExcepti
 @SuppressWarnings("unused")
 public class Session {
     // Only one session can run at a time so globally cache the current Session ID
-    private static String ID;
+    private static volatile String ID;
 
     private final String id;
 
@@ -38,7 +38,7 @@ public class Session {
         return new Session(Session.ID);
     }
 
-    public synchronized static void deleteGlobalSession() {
+    public static void deleteGlobalSession() {
         Session.ID = null;
     }
 }
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Session.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Session.java
@@ -30,7 +30,7 @@ public class Session {
      * @return Serializable Session object
      * @throws SessionNotCreatedException Thrown if a Session is already running
      */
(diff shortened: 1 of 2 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Session.java
```diff
@@ -2,31 +2,42 @@
...
     public static String getGlobalSessionId() {
         return Session.ID;
     }
 
+    /**
+     * Create a global session. Only one session can run per server instance so throw an exception if one already is in progress
+     * @return Serializable Session object
+     * @throws SessionNotCreatedException Thrown if a Session is already running
+     */
+    public static Session createGlobalSession() throws SessionNotCreatedException {
```
mykola-mokhnach: I assume you are missing **synchronized** modifier for this method

dpgraham: Yeah I should add that.

Is that standard practice when dealing with static variables?

mykola-mokhnach: this is standard practice when dealing with shared data inside methods (no matter these are static or not).  synchronized is like a semaphore, which guards the whole method, so it's not recommended to use it for big methods. For such cases one may use targeted locks (ReentrantLock), for example.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Session.java
```diff
@@ -2,32 +2,43 @@
...
+     */
+    public synchronized static Session createGlobalSession() throws SessionNotCreatedException {
+        if (Session.ID != null) {
+            throw new SessionNotCreatedException(String.format("Session %s is already in progress. Appium Espresso Server can only handle one session at a time.", Session.ID));
+        }
+        Session.ID = UUID.randomUUID().toString();
+        return new Session(Session.ID);
+    }
+
+    public synchronized static void deleteGlobalSession() {
```
mykola-mokhnach: there's no need to make this method synchronized

mykola-mokhnach: also, in case you only have a single variable there then you can add **volatile** modifier to its definition to make sure it is safe to access it from multiple threads.

dpgraham:
Changes applied.

PR #38, opened 2017-06-01
Save desiredCapabilities with global session

dpgraham:
* Instead of a static global session ID have a static instance of Session which contains the ID and the desiredCapabilities
* Refactored Session and other methods that use Session to use this new format
* Added routes GET /session/:sessionId and GET /sessions
* Fixed bug in RouteMap; wasn't narrowing down route matches by method

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetSessions.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetSessions.java
@@ -1,17 +1,26 @@
 package io.appium.espressoserver.lib.handlers;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
 import javax.annotation.Nullable;
 
 import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
 import io.appium.espressoserver.lib.model.AppiumParams;
 import io.appium.espressoserver.lib.model.Session;
 
-public class GetSessions implements RequestHandler<AppiumParams, String[]> {
+public class GetSessions implements RequestHandler<AppiumParams, Collection<String>> {
 
     @Override
     @Nullable
-    public String[] handle(AppiumParams params) throws AppiumException {
-        return new String[]{ Session.getGlobalSession().getId() };
+    public Collection<String> handle(AppiumParams params) throws AppiumException {
+        // Returns all of
+        List<String> list = new ArrayList<>();
+        if(Session.getGlobalSession() != null)
+            list.add(Session.getGlobalSession().getId());
+        return Collections.unmodifiableList(list);
     }
 
 }
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetSessions.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetSessions.java
@@ -1,26 +1,24 @@
 package io.appium.espressoserver.lib.handlers;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
+import static java.util.Collections.unmodifiableList;
+import static java.util.Collections.singletonList;
 
 import javax.annotation.Nullable;
 
 import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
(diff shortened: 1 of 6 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/RouteMap.java
```diff
@@ -21,13 +21,12 @@ public void addRoute(RouteDefinition route) {
...
     @Nullable
-    public RouteDefinition findMatchingRoute(String uri) {
-        for (Map.Entry<Method, Map<String, RouteDefinition>> methodMap: routeMap.entrySet()) {
-           for (Map.Entry<String, RouteDefinition> route: methodMap.getValue().entrySet()) {
-               RouteDefinition routeDefinition = route.getValue();
-               if (routeDefinition.isMatch(uri)) {
-                   return routeDefinition;
-               }
+    public RouteDefinition findMatchingRoute(Method method, String uri) {
+        Map<String, RouteDefinition> methodMap = routeMap.get(method);
```
mykola-mokhnach: should we check if the key exists in the map first to avoid unexpected exception??

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetSessions.java
```diff
@@ -0,0 +1,17 @@
...
+
+import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.model.AppiumParams;
+import io.appium.espressoserver.lib.model.Session;
+
+public class GetSessions implements RequestHandler<AppiumParams, String[]> {
+
+    @Override
+    @Nullable
+    public String[] handle(AppiumParams params) throws AppiumException {
```
mykola-mokhnach: I'd prefer Collections.unmodifiableList to "classic" array

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/RouteMap.java
dpgraham: Yes! Good eye.

dpgraham:
Changes made

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetSessions.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class GetSessions implements RequestHandler<AppiumParams, Collection<String>> {
+
+    @Override
+    @Nullable
+    public Collection<String> handle(AppiumParams params) throws AppiumException {
+        // Returns all of
+        List<String> list = new ArrayList<>();
+        if(Session.getGlobalSession() != null)
+            list.add(Session.getGlobalSession().getId());
```
mykola-mokhnach: return Collections.unmodifiableList(Collections.singletonList(Session.getGlobalSession().getId()));

I assume it makes sense to move such wrapper into base class, because it's going to be used more than one time

mykola-mokhnach: It is also useful to use static imports for such Collections methods to make the code more readable.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetSessions.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.model.AppiumParams;
+import io.appium.espressoserver.lib.model.Session;
+
+public class GetSessions implements RequestHandler<AppiumParams, Collection<String>> {
```
mykola-mokhnach: Do we want to return sorted or unsorted sequence?
If sorted then it's better to use List interface for the returned type.

dpgraham: This server can only ever run one session at a time so we don't need to worry about sorting.

dpgraham:
Changes made

PR #42, opened 2017-06-02
Move /lib files over to main so can be accessed in unit tests

dpgraham:
* When src files are in androidTest/ they can't be accessed by test/
* When src files are in main/ they can be accessed by androidTest/ and test/
* Moved the files back over to main (main still has no Android app so there still won't be an overhead of running an Android app inside of devices)
* Added a unit Test for RouteDefinition.java
* Doing this because I'll need to write unit tests to for the upcoming /source implementation


dpgraham:
Closing until I resolve Travis CI

PR #44, opened 2017-06-06
Added /source handler

dpgraham:
* Added RootViewFinder which trickily implements ViewAction to get a reference to the Root view in the app
* Added source handler endpoint that passes the Root View into the ElementNode and serializes XML that is returned to the client
* Made use of javax.xml to build the XML
* NOTE: I attempted to use simplexml and tried some other XML parsing libraries and they were all very problematic (SimpleXML I couldn't rename the tags dynamically). Since this is the only place that we'll use XML I think using javax.xml should be fine
* Added a mocha test to confirm that it can get the source, parse it's XML and then query it using XPath


dpgraham:
@jlipps Yeah I was thinking I would extract this out to it's own class and then during the recursive visits to all the View's, cache the Views in a Map and then having a method that applies the XPath on the document and then locates the cached view.

dpgraham:
Changes made. It adds a bunch more attributes, and makes note of the attributes that still need to be set. Follows same format as android-driver

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    @Nullable
+    public String handle(AppiumParams params) throws AppiumException {
+
+        try {
+            // Build a document
+            Document doc;
+            DocumentBuilderFactory docFactory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder docBuilder = docFactory.newDocumentBuilder();
+            doc = docBuilder.newDocument();
```
mykola-mokhnach: I suppose all these factory methods and stuff they produce could be cached. There is no need to call them for each invocation.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
```diff
@@ -0,0 +1,100 @@
...
+
+            // Populate the XML
+            View rootView = (new RootViewFinder()).getRootView();
+            buildXML(doc, null, rootView);
+
+            // Write the XML to string
+            TransformerFactory transformerFactory = TransformerFactory.newInstance();
```
mykola-mokhnach: Can we also add some simple formatting for the resulting xml document on this stage?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    /**
+     * Recursively visit all of the views and map them to XML elements
+     * @param doc XML Document
+     * @param parentElement Element that this new element will be appended to
+     * @param view Android View that will map to an Element
+     */
+    public void buildXML(Document doc, Element parentElement, View view) {
```
mykola-mokhnach: private?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
```diff
@@ -0,0 +1,100 @@
...
+        // TODO: Extract this logic out to it's own class so it can be re-used with XPath locators
+        Element element = doc.createElement(view.getClass().getName());
+
+        // Set attributes
+        ViewElement viewElement = new ViewElement(view);
+        element.setAttribute(ViewAttributesEnum.CONTENT_DESC.getName(), viewElement.getContentDescription());
```
mykola-mokhnach: I find this is not very safe, since XML standard explicitly defines the set of acceptable characters. It would be better to have helper methods, which verifies the input and replaces unacceptable characters with empty chars.
Hint: you can copy such method from uiautomator2-server source (the name is **stripInvalidXMLChars**).

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
```diff
@@ -0,0 +1,100 @@
...
+        element.setAttribute(ViewAttributesEnum.BOUNDS.getName(), viewElement.getBounds());
+        element.setAttribute(ViewAttributesEnum.FOCUSED.getName(), viewElement.getFocused());
+        element.setAttribute(ViewAttributesEnum.CLICKABLE.getName(), viewElement.getClickable());
+        element.setAttribute(ViewAttributesEnum.LONG_CLICKABLE.getName(), viewElement.getLongClickable());
+        element.setAttribute(ViewAttributesEnum.CLASS.getName(), viewElement.getClassName());
+        element.setAttribute(ViewAttributesEnum.INDEX.getName(), viewElement.getIndex());
```
mykola-mokhnach: can we add more properties here? The _visible_ one would be super useful, for example.
UIA2 defines, two times more properties, for example (see **dumpNodeRec** method)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.java
```diff
@@ -24,12 +24,12 @@ public Server() throws IOException, DuplicateRouteException {
         router = new Router();
     }
 
-    private String[] getStackTrace(Exception e) {
+    private ArrayList<String> getStackTrace(Exception e) {
```
mykola-mokhnach: It is recommended to always use abstract/interface types instead of concrete types (`ArrayList<String>` -> `List<String>` )

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.java
```diff
@@ -38,8 +38,8 @@ public Response serve(IHTTPSession session) {
         BaseResponse response;
         try {
             response = router.route(session);
-        } catch (Exception e) {
-            String[] stackTrace = getStackTrace(e);
+        } catch (RuntimeException e) {
+            ArrayList<String> stackTrace = getStackTrace(e);
```
mykola-mokhnach: -> List<String>

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ViewAttributesEnum.java
```diff
@@ -0,0 +1,34 @@
...
+    LONG_CLICKABLE("long-clickable"),
+    PASSWORD("password"),
+    SELECTED("selected"),
+    BOUNDS("bounds"),
+    RESOURCE_ID("resource-id"),
+    INSTANCE("instance"),
+    INDEX("index");
+
+
+    private final String name;
```
mykola-mokhnach: each enum element already has **name** property. We could simply transform it to lowecase, replace underscore with dash and override **toString** to avoid having extra constructor.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ViewElement.java
```diff
@@ -0,0 +1,104 @@
...
+        // scrollable	false
+        // password	false
+        // selected	false
+        // bounds	[0,0][1440,2560]
+        // resource-id	android:id/decor_content_parent
+        // instance	0
+
+    }
+
+    public String getContentDescription() {
```
mykola-mokhnach: I'd not return strings here, but rather "original" types. We can convert them inside xml writer if we need it. In the worst case one my just box simple types like boolean or int into Boolean/Integer and then call toString().toLowerCase(), where needed.

There will be another API calls like isClickable, which should return boolean value, so we will have to perform "reverse" type conversation again there

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/viewaction/RootViewFinder.java
```diff
@@ -0,0 +1,45 @@
+package io.appium.espressoserver.lib.viewaction;
+
+import android.support.test.espresso.UiController;
+import android.support.test.espresso.ViewAction;
+import android.support.test.espresso.matcher.ViewMatchers;
+import android.view.View;
+
+import org.hamcrest.Matcher;
```
mykola-mokhnach: Is this import used?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/viewaction/RootViewFinder.java
```diff
@@ -0,0 +1,45 @@
...
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.matcher.ViewMatchers.isRoot;
+
+/**
+ * Get the Root View of the Android App
+ */
+public class RootViewFinder {
```
mykola-mokhnach: this class implementation is quite hacky and cryptic. Would it be possible to add more documentation here?

test/functional/source-e2e-specs.js
```diff
@@ -0,0 +1,41 @@
+import chai from 'chai';
```
mykola-mokhnach: 👍

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ViewElement.java
```diff
@@ -0,0 +1,104 @@
...
+public class ViewElement {
+
+    private final int id;
+    private final boolean clickable;
+    private final boolean longClickable;
+    private final boolean focused;
+    private final String className;
+    private int index;
+    private String contentDescription = "";
+    private String bounds;
```
mykola-mokhnach: can we represent bounds as Rect type instead?

mykola-mokhnach:
@dpgraham Try android.util.Xml package to generate result XML. There is a nice example of page source generation in UIA2 source. Check the **AccessibilityNodeInfoDumper** class

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
dpgraham: I'll make fixes for Source.java in the other PR I have open.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
```diff
@@ -66,17 +66,17 @@ public String handle(AppiumParams params) throws AppiumException {
...
         // Set attributes
         ViewElement viewElement = new ViewElement(view);
-        element.setAttribute(ViewAttributesEnum.CONTENT_DESC.getName(), viewElement.getContentDescription());
-        element.setAttribute(ViewAttributesEnum.BOUNDS.getName(), viewElement.getBounds());
-        element.setAttribute(ViewAttributesEnum.FOCUSED.getName(), viewElement.getFocused());
-        element.setAttribute(ViewAttributesEnum.CLICKABLE.getName(), viewElement.getClickable());
-        element.setAttribute(ViewAttributesEnum.LONG_CLICKABLE.getName(), viewElement.getLongClickable());
+        element.setAttribute(ViewAttributesEnum.CONTENT_DESC.getName(), viewElement.getContentDescription().toString());
+        element.setAttribute(ViewAttributesEnum.BOUNDS.getName(), viewElement.getBounds().toShortString());
+        element.setAttribute(ViewAttributesEnum.FOCUSED.getName(), viewElement.isFocused() ? "true" : "false");
```
mykola-mokhnach: we can also use toString here if we return Boolean type. Then we'll have less code duplication

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ViewElement.java
```diff
@@ -67,31 +68,32 @@ public ViewElement(View view) {
...
+    public CharSequence getContentDescription() {
         return contentDescription;
     }
 
-    public String getBounds() {
+    public Rect getBounds() {
         return bounds;
     }
 
     public String getId() {
```
mykola-mokhnach: why not int?

dpgraham: The bounds have four numbers: top, left, bottom, right

mykola-mokhnach: I mean the return type of getId method. How it is related to bounds?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ViewElement.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public boolean isFocused() {
+        return focused;
+    }
+
+    public String getClassName() {
+        return className;
+    }
+
+    public String getIndex() {
```
mykola-mokhnach: can getIndex return int as well?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Source.java
```diff
@@ -0,0 +1,100 @@
...
+            TransformerFactory transformerFactory = TransformerFactory.newInstance();
+            Transformer transformer = transformerFactory.newTransformer();
+            DOMSource source = new DOMSource(doc);
+            StringWriter stringWriter = new StringWriter();
+            StreamResult result = new StreamResult(stringWriter);
+            transformer.transform(source, result);
+            return stringWriter.toString();
+
+        } catch (ParserConfigurationException pe) {
+            throw new AppiumException(String.format("Could not parse XML from source: %s", pe.getMessage()));
```
mykola-mokhnach: does it make a sense to keep the parent stacktrace by adding _cause_ argument to AppiumException constructor?

dpgraham: What's the difference between message and cause?

Does cause contain the stacktrace?

mykola-mokhnach: https://stackoverflow.com/questions/5800629/setting-exception-cause-in-java
https://docs.oracle.com/javase/tutorial/essential/exceptions/chained.html

I personally find _cause_ very useful for such cases.

dpgraham:
I'm going to close this because this PR is rebased on top of this one:

https://github.com/appium/appium-espresso-driver/pull/45

PR #46, opened 2017-06-08
Override different serve definition in nanohttpd

dpgraham:
* There appeared to be a race condition that was causing multiple routes to have the same POST body (Isaac had this problem during the hack week)
* This way appears to fix it and is cleaner too

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
@@ -124,10 +124,7 @@ class Router {
 
 
     @SuppressWarnings("unchecked")
-    public BaseResponse route(IHTTPSession session) {
-        String uri = session.getUri();
-        Method method = session.getMethod();
-
+    public BaseResponse route(String uri, Method method, Map<String, String> params,  Map<String, String> files) {
         // Look for a route that matches this URL
         RouteDefinition matchingRoute = routeMap.findMatchingRoute(method, uri);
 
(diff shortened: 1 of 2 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.java
```diff
@@ -21,7 +22,7 @@
     public Server() throws IOException, DuplicateRouteException {
         super(8080);
```
mykola-mokhnach: can we move port number to a constant?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.java
```diff
@@ -34,11 +35,11 @@ public Server() throws IOException, DuplicateRouteException {
     }
 
     @Override
-    public Response serve(IHTTPSession session) {
+    public Response serve(String uri, Method method, Map<String, String> headers, Map<String, String> parms, Map<String, String> files) {
```
mykola-mokhnach: typo -> params

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
```diff
@@ -143,13 +140,7 @@ public BaseResponse route(IHTTPSession session) {
 
         // Parse the appium params
         String postJson;
```
mykola-mokhnach: you can combine variable definition and initialization now

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.java
dpgraham: The typo is in the NanoHTTPD. This is overriding a method.

dpgraham:
Fixes made

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.java
mykola-mokhnach: I assume in strictly-typed Java world only argument types matter, but you can use your own names for override method arguments. 
https://stackoverflow.com/questions/9627139/overriding-methods-with-different-variable-names

PR #47, opened 2017-06-08
Added screenshot handler

dpgraham:
I've confirmed that this works. I ran it in Appium Desktop. 

Side note, the espresso driver does everything we need to run it in Appium Desktop now! (screenshot, source, find by xpath, click, send keys)

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
@@ -1,0 +1,43 @@
+package io.appium.espressoserver.lib.handlers;
+
+import android.graphics.Bitmap;
+import android.util.Base64;
+import android.view.View;
+
+import java.io.ByteArrayOutputStream;
+
+import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.model.AppiumParams;
+import io.appium.espressoserver.lib.viewaction.RootViewFinder;
+
+public class Screenshot implements RequestHandler<AppiumParams, String> {
+
+    @Override
+    public String handle(AppiumParams params) throws AppiumException {
+        return takeScreenshot();
+    }
+
+
+    private String takeScreenshot() throws AppiumException {
+        try {
+            // Create bitmap screen capture
+            View rootView = (new RootViewFinder()).getRootView();
+            rootView.setDrawingCacheEnabled(true);
+            Bitmap bitmapScreenCap = Bitmap.createBitmap(rootView.getDrawingCache());
+            rootView.setDrawingCacheEnabled(false);
+
+            // Stream the bitmap to byte array stream
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+            int quality = 100;
+            bitmapScreenCap.compress(Bitmap.CompressFormat.JPEG, quality, outputStream);
+            byte[] bytes = outputStream.toByteArray();
+
+            // Encode the byte array stream to base 64
+            return Base64.encodeToString(bytes, Base64.DEFAULT);
+
+        } catch (Exception e) {
+            throw new AppiumException(String.format("Could not get screenshot %s", e.getCause()));
+        }
+    }
+
+}
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
@@ -2,19 +2,16 @@ package io.appium.espressoserver.lib.http;
 
 import com.google.gson.Gson;
 
-import java.io.IOException;
-
-import java.util.HashMap;
 import java.util.Map;
 
 import fi.iki.elonen.NanoHTTPD;
 import fi.iki.elonen.NanoHTTPD.Method;
-import fi.iki.elonen.NanoHTTPD.IHTTPSession;
 import io.appium.espressoserver.lib.handlers.Click;
 import io.appium.espressoserver.lib.handlers.CreateSession;
 import io.appium.espressoserver.lib.handlers.NotYetImplemented;
 import io.appium.espressoserver.lib.handlers.GetSession;
 import io.appium.espressoserver.lib.handlers.GetSessions;
+import io.appium.espressoserver.lib.handlers.Screenshot;
 import io.appium.espressoserver.lib.handlers.Source;
 import io.appium.espressoserver.lib.handlers.exceptions.InvalidStrategyException;
 import io.appium.espressoserver.lib.handlers.exceptions.NotYetImplementedException;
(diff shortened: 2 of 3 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private String encodeFileToBase64Binary(File file) throws IOException {
+        byte[] bytes = loadFile(file);
+        byte[] encoded = Base64.encode(bytes, 0);
+        String encodedString = new String(encoded);
+
+        return encodedString;
+    }
+
+    private static byte[] loadFile(File file) throws IOException {
```
jlipps: you gotta be kidding me that this is the easiest way to get a byte array out of a file!

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+
+    private String takeScreenshot() throws AppiumException {
+        Date now = new Date();
+        android.text.format.DateFormat.format("yyyy-MM-dd_hh:mm:ss", now);
+
+        try {
+            // image naming and path  to include sd card  appending name you choose for file
+            String mPath = Environment.getExternalStorageDirectory().toString() + "/" + now + ".jpg";
```
mykola-mokhnach: as for me current date is not the best thing to be used for unique file identification. I'd rather use UUID class. We can even get only the first 8 chars from the generated id, - their uniqueness is enough.

mykola-mokhnach: can we set better name to this variable, for example targetScreenshotPath?

mykola-mokhnach: in Java world it's more useful to use File type for paths and other stuff. This way one can define it as:

```java
File screenshot = new File(Environment.getExternalStorageDirectory().toString(), fileName + ".jpg")
```

path separator will be added automatically

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+
+            // create bitmap screen capture
+            View v1 = (new RootViewFinder()).getRootView();
```
mykola-mokhnach: better name?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+            v1.setDrawingCacheEnabled(true);
+            Bitmap bitmap = Bitmap.createBitmap(v1.getDrawingCache());
```
mykola-mokhnach: better name?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+
+
+            FileOutputStream outputStream = new FileOutputStream(imageFile);
+            int quality = 100;
+            bitmap.compress(Bitmap.CompressFormat.JPEG, quality, outputStream);
+
+            outputStream.flush();
+            outputStream.close();
+            return encodeFileToBase64Binary(imageFile);
+        } catch (Throwable e) {
```
mykola-mokhnach: Newer catch **Throwable** explicitly, since it includes OutOfMemory exception and other ones, that we don't want to intercept for sure. In general case it is enough to catch just **Exception**.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+            // Several error may come out with file handling or OOM
+            e.printStackTrace();
+            throw new AppiumException(String.format("Could not get screenshot %s", e.getMessage()));
```
mykola-mokhnach: passing cause would be better here than printing the stack trace

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    private String encodeFileToBase64Binary(File file) throws IOException {
+        byte[] bytes = loadFile(file);
+        byte[] encoded = Base64.encode(bytes, 0);
+        String encodedString = new String(encoded);
```
mykola-mokhnach: there is no need to have this intermediate variable here

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
mykola-mokhnach: natively yes. However we could also use https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#toByteArray(java.io.InputStream)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
```diff
@@ -193,11 +185,12 @@ public BaseResponse route(IHTTPSession session) {
         }
     }
 
-    private static String parseBody (IHTTPSession session) throws IOException, NanoHTTPD.ResponseException {
-        String result;
+    private static synchronized String parseBody (IHTTPSession session) throws IOException, NanoHTTPD.ResponseException {
```
mykola-mokhnach: what is the purpose of the is method?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
```diff
@@ -124,10 +125,7 @@
 
 
     @SuppressWarnings("unchecked")
-    public BaseResponse route(IHTTPSession session) {
-        String uri = session.getUri();
-        Method method = session.getMethod();
-
+    public synchronized BaseResponse route(String uri, Method method, Map<String, String> parms,  Map<String, String> files) {
```
mykola-mokhnach: having synchronized methods is usually expensive for performance. ReentrantLock is much better alternative to it. One just have to check carefully where to guard shared resources

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class Screenshot implements RequestHandler<AppiumParams, String> {
+
+    @Override
+    public String handle(AppiumParams params) throws AppiumException {
+        return takeScreenshot();
+    }
+
+
+    private String takeScreenshot() throws AppiumException {
```
mykola-mokhnach: I feel this method is doing some extra work. Cannot we just save the screenshot to a string/memory stream and do all the transformations in memory instead of creating a file and asking android.permission.WRITE_EXTERNAL_STORAGE permissions?

dpgraham:
Fixes pushed. @mykola-mokhnach I followed your recommendation and now it doesn't deal with the filesystem. It just streams the bitmap to a Byte Array and then encodes the byte array as Base64. Much simpler. Good suggestion.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Screenshot.java
```diff
@@ -0,0 +1,47 @@
...
+        Date now = new Date();
+        android.text.format.DateFormat.format("yyyy-MM-dd_hh:mm:ss", now);
```
mykola-mokhnach: is this used anywhere?

dpgraham: No it isn't needed any more. I removed it.

PR #56, opened 2017-06-16
Throws exception if element is stale

dpgraham:
* When calling Element.getById, the Element class checks if the view is still displayed
* If view is not displayed throw StaleElementException

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/StaleElementException.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/StaleElementException.java
@@ -1,0 +1,7 @@
+package io.appium.espressoserver.lib.handlers.exceptions;
+
+public class StaleElementException extends AppiumException {
+    public StaleElementException(String elementId) {
+        super(String.format("Element %s no longer exists", elementId));
+    }
+}
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/StaleElementException.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/StaleElementException.java
@@ -2,6 +2,6 @@ package io.appium.espressoserver.lib.handlers.exceptions;
 
 public class StaleElementException extends AppiumException {
     public StaleElementException(String elementId) {
-        super(String.format("Element %s no longer exists", elementId));
+        super(String.format("The cached element %s no longer exists in the Android View hierarchy. Try to find it using a locator.", elementId));
     }
 }
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
@@ -27,6 +27,7 @@ import io.appium.espressoserver.lib.handlers.RequestHandler;
 import io.appium.espressoserver.lib.handlers.DeleteSession;
 import io.appium.espressoserver.lib.handlers.SendKeys;
 import io.appium.espressoserver.lib.handlers.Status;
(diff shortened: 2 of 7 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/StaleElementException.java
```diff
@@ -0,0 +1,7 @@
+package io.appium.espressoserver.lib.handlers.exceptions;
+
+public class StaleElementException extends AppiumException {
+    public StaleElementException(String elementId) {
+        super(String.format("Element %s no longer exists", elementId));
```
mykola-mokhnach: I observe that people usually don't understand the meaning of this error and have no idea how to fix it. What if we put more details into exception message, for example:

``` 
The cached element %s no longer exists in the accessibility objects hierarchy. Try to find it using its locator.
```

dpgraham: I like that. I would change it to View hierarchy, maybe? Correct me if I'm wrong, but accessibility objects are more of a UIAutomator thing right?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
```diff
@@ -178,6 +179,8 @@ public BaseResponse route(String uri, Method method, Map<String, String> params,
             return new ErrorResponse(NanoHTTPD.Response.Status.NOT_FOUND, e.getMessage());
         } catch (NotYetImplementedException e) {
             return new ErrorResponse(NanoHTTPD.Response.Status.NOT_IMPLEMENTED, e.getMessage());
+        } catch (StaleElementException e) {
+            return new AppiumResponse<>(AppiumStatus.STALE_ELEMENT_REFERENCE, e.getMessage());
```
mykola-mokhnach: do we want to keep only message or the whole exception cause? I would prefer the second one

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/StaleElementException.java
mykola-mokhnach: np, this was just an example.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
dpgraham: Not sure if we have to or not. I put it in there because it's part of the spec. @jlipps do we need to keep the MJSONWP error messages or can we use our own?

mykola-mokhnach: probably, we could add  e.printStackTrace() instead to simplify problems investigation in logcat output

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Router.java
```diff
@@ -169,16 +170,25 @@ public BaseResponse route(String uri, Method method, Map<String, String> params,
             System.out.println("Finished processing " + method + " request for '" + uri + "'");
             return appiumResponse;
         } catch (NoSuchElementException e) {
+            e.printStackTrace();
```
jlipps: the duplication here is a bit intense. could we create a method like:

```java
protected AppiumResponse errorResponse (e, AppiumStatus status) {
  e.printStackTrace();
  return new AppiumResponse<>(status, e.getMessage());
}
```

then in each of these catches we could just do `return errorResponse(e, AppiumStatus.NO_SUCH_ELEMENT);`

?

dpgraham: Yeah, good call. I just pushed an update for that.

PR #74, opened 2017-12-23
add gradle build files to gitignore, there's no need for this files(98% sure) 

vrunoa:
and it's generating the module to be more than 100MB


vrunoa:
🎉

PR #75, opened 2017-12-24
Backport W3C actions implementation for UIA2 server

mykola-mokhnach:
Also updated gradle dependencies.


mykola-mokhnach:
I fixed what I could from Codacy comments. The rest requires too much effort, which is overkill for a product in beta state. I'd rather merge it as is and let testers to try it.

dpgraham:
Sure that's fine with me.

PR #91, opened 2018-02-28
Add handler for `/session/:sessionId/element/active` endpoint

mykola-mokhnach:
There are both W3C and non-W3C routes

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/FindActive.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/FindActive.java
@@ -16,21 +16,22 @@
 
 package io.appium.espressoserver.lib.handlers;
 
+import android.support.test.espresso.ViewInteraction;
+
 import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.handlers.exceptions.NoSuchElementException;
 import io.appium.espressoserver.lib.model.AppiumParams;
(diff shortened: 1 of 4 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/FindActive.java
```diff
@@ -0,0 +1,36 @@
...
+import io.appium.espressoserver.lib.model.ViewAttributesEnum;
+
+public class FindActive implements RequestHandler<AppiumParams, Element> {
+
+    @Override
+    public Element handle(AppiumParams params) throws AppiumException {
+        final Locator locator = new Locator();
+        locator.initUriMapping(params);
+        locator.setUsing(Strategy.XPATH);
+        locator.setValue(String.format("//*[@%s='true']", ViewAttributesEnum.FOCUSED.toString()));
```
vmaxim: Maybe it's better to use  `View.isFocused()` instead this trick?

mykola-mokhnach: It would generate much more code than this

vmaxim: Few extra lines better than magic locators. Also performance will be increased significantly.

mykola-mokhnach:
@vmaxim how do you find the updated implementation?

vmaxim:
@mykola-mokhnach Nice. Easy to read, easy to understand and pure espresso.

PR #94, opened 2018-03-07
Improve error handling

mykola-mokhnach:
This makes thrown exceptions more specific and helps the server in general to be more compatible to what JSWOWP spec requires.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/AcceptAlert.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/AcceptAlert.java
@@ -28,6 +28,7 @@ import java.util.List;
 import javax.annotation.Nullable;
 
 import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.handlers.exceptions.InvalidElementStateException;
 import io.appium.espressoserver.lib.handlers.exceptions.NoAlertOpenException;
 import io.appium.espressoserver.lib.helpers.Logger;
 import io.appium.espressoserver.lib.model.AppiumParams;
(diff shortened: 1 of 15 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/AcceptAlert.java
```diff
@@ -42,7 +43,7 @@ public Void handle(AppiumParams params) throws AppiumException {
                 .getInstance(InstrumentationRegistry.getInstrumentation());
         final List<UiObject2> dialogs = mDevice.findObjects(By.clazz(Dialog.class));
         if (dialogs.isEmpty()) {
-            throw new AppiumException("No alerts can be detected on the screen");
+            throw new NoAlertOpenException("No alerts can be detected on the screen");
```
vmaxim: ```"No alerts can be detected on the screen"``` string can be moved to the  ```NoAlertOpenException``` class and used in the default constructor.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Clear.java
```diff
@@ -36,7 +37,7 @@ public Void handle(AppiumParams params) throws AppiumException {
         try {
             viewInteraction.perform(clearText());
         } catch (PerformException e) {
-            throw new AppiumException(String.format("Could not apply clear to element %s",
+            throw new InvalidElementStateException(String.format("Could not apply clear to element %s",
```
vmaxim: I think it would be better to implement a new constructor ```InvalidElementStateException(String command, String elementId)```.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/AcceptAlert.java
mykola-mokhnach: Changed

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Clear.java
mykola-mokhnach: these error messages are different for different cases, so I don't think such constructor may help to reduce LOC

vmaxim: ```"Could not apply clear to element %s"```
```"Could not click element %s"```
```"Could not apply sendKeys to element %s"```
=>
```"Could not perform %s action on element %s"```

PR #106, opened 2018-04-04
createSession return session's capabilities

theoden9014:
https://github.com/appium/appium-uiautomator2-driver/pull/157 for espresso.


dpgraham:
This seems to has caused some other problems in other drivers. Closing this.

PR #108, opened 2018-05-11
Get the basic driver tests running

imurchie:
A variety of things, to get the driver running:
* Update gradle
* Massage dependencies to run on x86 and ARM emulators (so Travis works)
* Rename `Finder` class to `FindElement`, so it is in line with everywhere else
* Removed some superfluous tests

The Travis config now _only runs the driver tests_. The command tests fail for various reasons, locally and on Travis, and can be added to the config as they get fixed.


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetDisplayed.java
```diff
@@ -30,11 +32,10 @@
 
     @Override
     public Boolean handle(AppiumParams params) throws AppiumException {
-        final ViewInteraction viewInteraction = Element.getById(params.getElementId());
         try {
-            viewInteraction.check(matches(isDisplayed()));
+            Element.getById(params.getElementId());
```
mykola-mokhnach: why so?

lib/driver.js
```diff
@@ -321,7 +331,11 @@ class EspressoDriver extends BaseDriver {
     // TODO below logic is duplicated from uiautomator2
     if (this.adb) {
       if (this.wasAnimationEnabled) {
-        await this.adb.setAnimationState(true);
+        try {
+          await this.adb.setAnimationState(true);
+        } catch (err) {
+          logger.debug(`Unable to reset animation: ${err.message}`)
```
mykola-mokhnach: I'd rather warn

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetDisplayed.java
imurchie: Because `Element.getById` only returns an element if the `viewInteraction.check(matches(isDisplayed()));` has passed. Otherwise it throws a `StaleElementException`.

mykola-mokhnach: does it always have an id?

imurchie: As much as it ever did. If it doesn't `Element.getById` will throw a `NoSuchElementException` as it did before.

PR #111, opened 2018-05-15
Implement an endpoint for pressing keys

mykola-mokhnach:
The actual implementation was backported from UIA2 server


mykola-mokhnach:
We'd need to have a separate java library to keep the common code there. I don't know who is going to maintain and publish it.

jlipps:
Well currently you are the one copying and pasting code, so as long as you're fine continuing to do that... :-p

PR #112, opened 2018-05-17
Cache elements by view instead of view interaction

dpgraham:
* Updated the element finding logic so that instead of getting View Interactions it gets the actual Views
* Updated the Element.java class to cache the actual views
* The reasoning for this is that it appears that viewInteractions become stale after using them once, so you can't operate on an "Element" more than once
* @jlipps I think  this may solve your issue with finding elements too

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
@@ -52,7 +52,7 @@ import static org.hamcrest.Matchers.endsWith;
 public class ViewFinder {
 
     @Nullable
-    public static ViewInteraction findBy(Strategy strategy, String selector)
+    public static View findBy(Strategy strategy, String selector)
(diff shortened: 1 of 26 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
```diff
@@ -143,57 +143,61 @@ public void describeTo(Description description) {
...
+                    interaction = onView(withIndex(matcher, 0));
+                } else {
+                    interaction = onView(withIndex(matcher, 0)).inRoot(withDecorView(is(root)));
                 }
-                return Collections.singletonList(
-                        onView(withIndex(matcher, 0)).inRoot(withDecorView(is(root)))
-                );
+                View view = (new io.appium.espressoserver.lib.viewaction.ViewFinder()).getView(interaction);
+                return Collections.singletonList(view);
             } catch (NoMatchingViewException e) {
```
mykola-mokhnach: isn't this a subclass of Exception?

PR #120, opened 2018-05-23
Make accessibility selector scroll into view

dpgraham:
* When locating an element by accessibility ID, if it isn't found on the first time, try using onData to see if it can be matched with a Hamcrest matcher and scrolled into view (see https://medium.com/google-developers/adapterviews-and-espresso-f4172aa853cf for info on how `onData` works)
* Added `adapter-type` and `adapters` as an attribute to source. This was from a previous experimental branch I worked on. I think in the future I'd like to use a Hamcrest matcher strategy for locating elements so we have other ways to target elements that aren't visible
* Added a test for finding an element that is off the screen (has an accessibility ID called WebView which is the bottom of the list)

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
@@ -171,17 +171,8 @@ public class ViewFinder {
 
                 // If the item is not found on the screen, use 'onData' to try
                 // to scroll it into view and then locate it again
-                if (views.isEmpty()) {
-                    try {
-                        DataInteraction dataInteraction = onData(
-                                hasEntry(Matchers.equalTo("contentDescription"), is(selector))
-                        );
-                        dataInteraction.check(matches(isDisplayed()));
-                    } catch (Exception e) {
-                        // Didn't find it. Don't do anything
-                    } finally {
-                        views = getViews(root, withContentDescription(selector), findOne);
-                    }
+                if (views.isEmpty() && canScrollToViewWithContentDescription(selector)) {
+                    views = getViews(root, withContentDescription(selector), findOne);
                 }
                 break;
             case XPATH:
@@ -198,6 +189,24 @@ public class ViewFinder {
         }
 
         return views;
+    }
+
+    /**
+     * Attempts to scroll to a view with the content description using onData
+     * @param contentDesc Content description
+     * @return
+     */
+    private static boolean canScrollToViewWithContentDescription (String contentDesc) {
+        try {
+            DataInteraction dataInteraction = onData(
+                    hasEntry(Matchers.equalTo("contentDescription"), is(contentDesc))
+            );
+            dataInteraction.check(matches(isDisplayed()));
+        } catch (Exception e) {
+            return false;
+        } finally {
+            return true;
+        }
     }
 
     private static List<View> getViews(
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
@@ -18,17 +18,21 @@ package io.appium.espressoserver.lib.helpers;
 
 import android.content.Context;
 import android.support.test.InstrumentationRegistry;
+import android.support.test.espresso.DataInteraction;
 import android.support.test.espresso.NoMatchingViewException;
+import android.support.test.espresso.ViewAssertion;
 import android.support.test.espresso.ViewInteraction;
 import android.view.View;
 
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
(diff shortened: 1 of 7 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
```diff
@@ -155,8 +167,22 @@ public void describeTo(Description description) {
                 views = getViews(root, withText(selector), findOne);
                 break;
             case ACCESSIBILITY_ID:
-                // with content description
                 views = getViews(root, withContentDescription(selector), findOne);
+
+                // If the item is not found on the screen, use 'onData' to try
```
mykola-mokhnach: why we only do it for elements located by id?

mykola-mokhnach: this introduces inconsistency.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
```diff
@@ -127,6 +134,46 @@ private void buildXML(Document doc, Element parentElement, View view) {
             parentElement.appendChild(element);
         }
 
+        // If it's an AdapterView, get the adapters as a String
```
mykola-mokhnach: I'd rather extract this into a separate method

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
```diff
@@ -127,6 +134,46 @@ private void buildXML(Document doc, Element parentElement, View view) {
...
+        try {
+            AdapterView adapterView = (AdapterView) view;
```
mykola-mokhnach: `if (view instanceof AdapterView)` ...

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
```diff
@@ -127,6 +134,46 @@ private void buildXML(Document doc, Element parentElement, View view) {
...
+            StringBuilder adapterData = new StringBuilder();
+            for (int i = 0; i < adapter.getCount(); i++) {
+                Object adapterItem = adapter.getItem(i);
+                adapterData.append(adapterItem);
+                if (i < adapter.getCount() - 1) {
+                    adapterData.append(",");
+                }
+
+                // Get the type of the adapter item
+                if (i == 0) {
```
mykola-mokhnach: `adapterItemType = adapterItem.getClass().getShortName()`
Also, it might be handy to apply the same class name filter that we apply for other classes

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
dpgraham: Only elements located by content description can be targeted with a data matcher. This is like an enhancement to the accessibility id selector: If the element with the Id isn't on screen, see if we can scroll to it

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
```diff
@@ -127,6 +134,27 @@ private void buildXML(Document doc, Element parentElement, View view) {
...
+                    setAttribute(element, ViewAttributesEnum.ADAPTER_TYPE, adapterItem.getClass().getSimpleName());
+                }
+                setAttribute(element, ViewAttributesEnum.ADAPTERS, adapterData);
+
```
mykola-mokhnach: extra line break here

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
```diff
@@ -127,6 +134,27 @@ private void buildXML(Document doc, Element parentElement, View view) {
...
+                adapterData.append(adapterItem);
+                if (i < adapter.getCount() - 1) {
+                    adapterData.append(",");
+                }
+
+                // Get the type of the adapter item
+                if (i == 0) {
+                    setAttribute(element, ViewAttributesEnum.ADAPTER_TYPE, adapterItem.getClass().getSimpleName());
+                }
+                setAttribute(element, ViewAttributesEnum.ADAPTERS, adapterData);
```
mykola-mokhnach: adapterData.toString()

dpgraham: It seems to have serialized it correctly without using `toString()`

mykola-mokhnach: yeah, I know, but explicit is always better than implicit

PR #129, opened 2018-05-26
add some no proxy commands

KazuCocoa:
I've added some commands in no-proxy to handle app.
Do you have any plan to implement some of them in native Java code like [start_activity]( https://github.com/appium/appium-espresso-driver/pull/119/files), btw?

---

link: https://github.com/appium/ruby_lib/issues/598#issuecomment-392232221

PR #119, opened 2018-05-23
startActivity

dpgraham:
* Added the `start_activity` endpoint
* Moved startActivity into a common helper method
* Fixed one of the tests. It was trying to open a non-existent activity

* NOTE: I don't think we'll ever be able to use `appPackage` in Espresso Driver. Espresso only deals with the App-under-test

(connected to: https://github.com/appium/appium/issues/8670)

PR #134, opened 2018-05-30
Fix: Re-insert start_activity router

dpgraham:


KazuCocoa:
Yes. This change was included in https://github.com/appium/appium-espresso-driver/pull/119/files and my comment in https://github.com/appium/appium-espresso-driver/pull/129 was against it.
(This PR will fix missing a change in #119 since something happened, it looks.)

dpgraham:
@jlipps Yeah I think it should be removed.... I'm pretty sure using `adb` start activity works but for how Espresso works, I don't think we want people trying to open activities that are outside of the A.U.T.

dpgraham:
Opened a new issue to have it removed: https://github.com/appium/appium-espresso-driver/issues/new

PR #165, opened 2018-06-13
Implement dispatcher algorithm 17.4

dpgraham:
(see https://www.w3.org/TR/webdriver/#dispatching-actions)

* Implements `dispatch tick actions`, `compute the tick duration` and `dispatch actions`

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/models/ActionSequence.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/models/ActionSequence.java
@@ -6,6 +6,7 @@ import java.util.List;
 
 import io.appium.espressoserver.lib.handlers.exceptions.InvalidArgumentException;
 import io.appium.espressoserver.lib.handlers.exceptions.NotYetImplementedException;
+import io.appium.espressoserver.lib.helpers.w3c.adapter.W3CActionAdapter;
 
 import static io.appium.espressoserver.lib.helpers.w3c.models.W3CActions.processSourceActionSequence;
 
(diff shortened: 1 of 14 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/models/ActionSequence.java
```diff
@@ -41,11 +51,55 @@ public ActionSequence(W3CActions w3CActions) throws InvalidArgumentException, No
...
+            if (!callables.isEmpty()) {
+                Executor executor = Executors.newFixedThreadPool(callables.size());
+                CompletionService<Void> completionService = new ExecutorCompletionService<>(executor);
+                for (Callable<Void> callable : callables) {
+                    completionService.submit(callable);
+                }
+
+                int received = 0;
+                while (received < callables.size()) {
+                    Future<Void> resultFuture = completionService.take(); //blocks if none available
```
mykola-mokhnach: CompltableFuture would be a better option

espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/TickTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+        int received = 0;
+        boolean errors = false;
+        while (received < callables.size() && !errors) {
+            Future<Void> resultFuture = completionService.take(); //blocks if none available
+            try {
+                resultFuture.get();
+                received ++;
+            }
+            catch(Exception e) {
```
mykola-mokhnach: Why exceptions should be caught in a test?

dpgraham: Good point, I could probably just not count the errors and let it throw the exception so that the test fails.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/models/ActionSequence.java
dpgraham: Do you have any code examples using `CompletableFuture`? I can't find anything that takes a list of callables and then translates it to a CompletableFuture

mykola-mokhnach: Perhaps, in this particular case we can leave with what we have, but in general I prefer to use CompletableFuture where possible, since it can do much much more that just a generic Future.

PR #166, opened 2018-06-13
Refactored processor functions

dpgraham:
* W3CActions was too bloated with all the processing functions so moved processors into separate classes
* Also removed `W3C` prefix from W3CActions and W3CActionsParams


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/models/ActionSequence.java
```diff
@@ -31,12 +31,14 @@
...
     private int tickCounter = 0;
 
-    public ActionSequence(W3CActions w3CActions) throws InvalidArgumentException, NotYetImplementedException {
+    public ActionSequence(Actions actions, ActiveInputSources activeInputSources,
+                          InputStateTable inputStateTable)
+            throws InvalidArgumentException, NotYetImplementedException {
         // Check if null to keep Codacy happy. It will never make it this far if it's null though.
-        if (w3CActions.getActions() != null) {
-            for (InputSource inputSource : w3CActions.getActions()) {
+        if (actions.getActions() != null) {
```
mykola-mokhnach: I'd rather use the built-in checkNotNull helper

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/processor/ActionsProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+                    actionObjects.add(processNullAction(action, inputSource.getType(), id, index));
+                    break;
+                case POINTER:
+                    actionObjects.add(processPointerAction(action, inputSource, id, index));
+                    break;
+                case KEY:
+                    actionObjects.add(processKeyAction(action, inputSource.getType(), id, index));
+                    break;
+                default:
+                    break;
```
mykola-mokhnach: should we throw instead?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/processor/PointerProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+    public static ActionObject processPointerMoveAction(Action action, InputSourceType inputSourceType, String id, int index) throws InvalidArgumentException {
+        ActionObject actionObject = new ActionObject(id, inputSourceType, action.getType(), index);
+
+        // 1-3 Add the duration
+        Long duration = action.getDuration();
+        assertNullOrPositive(index, id, "duration", duration);
+        actionObject.setDuration(duration);
+
+        // 4-7 Add the origin
+        String origin = action.getOrigin();
```
mykola-mokhnach: I'm not sure if origin is parsed properly if an element id is set there (like `{"element-6066-11e4-a52e-4f735466cecf": origin.id}`)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/processor/ActionsProcessor.java
dpgraham: Yeah. Good catch.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/processor/PointerProcessor.java
dpgraham: Yeah, I just took the raw string. If we were only parsing W3C I would look for the `element-` prefix but I think we need to support MJSONWP as well.

dpgraham: Aha, never mind I just thought it over again.... we need to expect the elements to be in the format `element-<element-id>`

mykola-mokhnach: not really. The `6066-11e4-a52e-4f735466cecf` suffix seems to be constant. See https://github.com/SeleniumHQ/selenium/commit/d5b792d1c0e6dd164512819854c462dd30c75a19 or https://github.com/SeleniumHQ/selenium/commit/573baceab793fff51f228aee400d2f7f36987671

dpgraham: Right, I got that confused with element id's....

So the format supposed to be: `{"origin": {"element-6066-11e4-a52e-4f735466cecf": "<element-uuid>}`?

mykola-mokhnach: yes

dpgraham: Alright, I updated it so that instead of 'Origin' being a string type I made an 'Origin' class that has a `type` property and an `elementId` property and made the JSON validator check if it's an object with that code, set the type to ELEMENT and set the elementId to the <element-uuid>. Otherwise, if it's a string, set the type and don't set elementId

PR #172, opened 2018-06-15
Switch to SubProcess for better control

### LINKED ISSUE #124: Running different apps in sequence doesn't work
Steps to Reproduce
* Run a session in one app
* Run a session in another app
* The second session will keep using the signed debug apk from the first session

The way I was able to work around it was wiping the data on the emulator

imurchie:
`SubProcess` allows us to have a handle on the running instrumentation process, while `exec` is meant to be fire-and-forget.

Finally fix #124 in Travis.

### DIFF
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -10,6 +10,7 @@ const TEST_APK_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'ap
 const TEST_MANIFEST_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'AndroidManifest-test.xml');
 const TEST_APK_PKG = 'io.appium.espressoserver.test';
 const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild'];
+const ESPRESSO_SERVER_LAUNCH_TIMEOUT = 30000;
 
 class EspressoRunner {
   constructor (opts = {}) {
@@ -23,6 +24,8 @@ class EspressoRunner {
     this.proxyReqRes = this.jwproxy.proxyReqRes.bind(this.jwproxy);
 
     this.modServerPath = path.resolve(this.tmpDir, `${TEST_APK_PKG}_${version}_${this.appPackage}.apk`);
+
+    this.serverLaunchTimeout = opts.serverLaunchTimeout || ESPRESSO_SERVER_LAUNCH_TIMEOUT;
   }
 
   async installTestApk () {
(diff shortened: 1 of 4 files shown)


lib/espresso-runner.js
```diff
@@ -47,10 +48,13 @@ class EspressoRunner {
     logger.info(`Repackaging espresso server for: '${this.appPackage}'`);
     let packageTmpDir = path.resolve(this.tmpDir, this.appPackage);
     let newManifestPath = path.resolve(this.tmpDir, 'AndroidManifest.xml');
+    if (await fs.exists(newManifestPath)) {
```
mykola-mokhnach: rimraf?

lib/espresso-runner.js
```diff
@@ -101,6 +115,8 @@ class EspressoRunner {
       logger.warn(`Did not get confirmation Espresso deleteSession worked; ` +
           `Error was: ${err}`);
     }
+
+    await this.instProcess.stop();
```
mykola-mokhnach: should we check if the process is running?

lib/espresso-runner.js
```diff
@@ -66,29 +70,39 @@ class EspressoRunner {
...
+      logger.debug(`[Instrumentation]${line.trim()}`);
+    });
+
+    await this.instProcess.start((stdout, stderr) => {
+      if (stdout.includes('io.appium.espressoserver.EspressoServerRunnerTest:')) {
+        return true;
+      }
+      if (stderr.includes('Exception')) {
+        throw new Error(stderr.trim());
+      }
```
mykola-mokhnach: What is going to happen if none of the conditions have been reached?

lib/espresso-runner.js
```diff
@@ -66,29 +70,39 @@ class EspressoRunner {
   }
 
   async startSession (caps) {
-    let cmd = ['am', 'instrument', '-w', '-e', 'debug', process.env.ESPRESSO_JAVA_DEBUG === 'true' ? 'true' : 'false',
+    let cmd = ['shell', 'am', 'instrument', '-w', '-e', 'debug', process.env.ESPRESSO_JAVA_DEBUG === 'true' ? 'true' : 'false',
```
mykola-mokhnach: It might be more readable to group some options, for example:
```
const cmd = [
  'shell'
  'am', 'instrument', 
  '-w',
   '-e', process.env.ESPRESSO_JAVA_DEBUG === 'true' ? 'true' : 'false'
...
]
```

lib/espresso-runner.js
jlipps: @mykola-mokhnach a timeout error will be thrown

PR #186, opened 2018-07-27
Run pointer move on main thread

dpgraham:
* Pointer move events were being called on a separate thread, which Espresso doesn't allow (must be main thread)
* Updated the logic so that the callable returns a class that has a `perform` method that gets called
* The perform method gets called on the main thread instead of on the thread

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/BaseDispatchResult.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/BaseDispatchResult.java
@@ -1,0 +1,23 @@
+package io.appium.espressoserver.lib.helpers.w3c.dispatcher;
+
+import java.util.concurrent.Callable;
+
+import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+
+public abstract class BaseDispatchResult {
+
+    private Callable<BaseDispatchResult> next;
+    public boolean hasNext() {
+        return next != null;
+    }
+
+    public abstract void perform() throws AppiumException;
+
+    public Callable<BaseDispatchResult> getNext() {
+        return next;
+    }
+
+    public void setNext(Callable<BaseDispatchResult> next) {
+        this.next = next;
+    }
+}
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/BaseDispatchResult.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/BaseDispatchResult.java
@@ -2,6 +2,8 @@ package io.appium.espressoserver.lib.helpers.w3c.dispatcher;
 
 import java.util.concurrent.Callable;
 
+import javax.annotation.Nullable;
+
 import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
 
 public abstract class BaseDispatchResult {
@@ -13,6 +15,7 @@ public abstract class BaseDispatchResult {
 
     public abstract void perform() throws AppiumException;
 
+    @Nullable
     public Callable<BaseDispatchResult> getNext() {
         return next;
     }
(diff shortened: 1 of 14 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/BaseDispatchResult.java
```diff
@@ -0,0 +1,23 @@
...
+public abstract class BaseDispatchResult {
+
+    private Callable<BaseDispatchResult> next;
+    public boolean hasNext() {
+        return next != null;
+    }
+
+    public abstract void perform() throws AppiumException;
+
+    public Callable<BaseDispatchResult> getNext() {
```
mykola-mokhnach: nullable

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/PointerDispatch.java
```diff
@@ -219,69 +219,82 @@ public static void dispatchPointerCancel(final W3CActionAdapter dispatcherAdapte
...
+
+                    // 8.3. Let input state's x property equal x and y property equal y
+                    pointerInputState.setX(x);
+                    pointerInputState.setY(y);
+                }
+
+                // Sleep for a fixed period of time
+                if (duration > 0) {
+                    try {
                         dispatcherAdapter.sleep(dispatcherAdapter.pointerMoveIntervalDuration());
```
mykola-mokhnach: it won't throw InterruptedException if we use SystemClock

espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/PointerDispatchTest.java
```diff
@@ -46,24 +47,26 @@
...
+        BaseDispatchResult dispatchResult = executorService.submit(callable).get();
+        dispatchResult.perform();
         executorService.shutdown();
         assertEquals(dummyW3CActionAdapter.getPointerMoveEvents().size(), 1);
     }
 
     @Test
-    public void shouldDoOneMoveIfDurationZero() throws ExecutionException, InterruptedException {
+    public void shouldDoOneMoveIfDurationZero()
+            throws ExecutionException, InterruptedException, AppiumException {
```
mykola-mokhnach: shouldn't ExecutionException/InterruptedException be covered into  AppiumException?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/PointerDispatch.java
dpgraham: I just changed it to throw an AppiumException and the Dummy Adapter used for testing catches the InterruptedException internally

espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/PointerDispatchTest.java
dpgraham: `ExecutionException` and `InterruptedException` are thrown by `executorService.submit(callable).get()`

mykola-mokhnach: I know, but we can transform them in try-catch

dpgraham: Does that matter for a unit test though?

mykola-mokhnach: for testing apparently not

dpgraham: I'm still trying to understand the finer points of Java...

I understand that in that we should be catching exceptions like InterruptedException and ExecutionException and throw them as AppiumException because the exceptions need to translate to proper webdriver errors.

In this case, though, it's just a unit test that doesn't get called by other methods, so does it make a difference what exceptions are thrown?

PR #189, opened 2018-07-30
viewtag selector strategy

dpgraham:
Add custom selector that targets elements by the [View Tag](https://developer.android.com/reference/android/view/View)

(related to: https://github.com/appium/appium/issues/6025)

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
@@ -47,11 +47,15 @@ import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static android.support.test.espresso.matcher.ViewMatchers.withClassName;
 import static android.support.test.espresso.matcher.ViewMatchers.withContentDescription;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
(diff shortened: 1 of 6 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ViewElement.java
```diff
@@ -248,4 +231,13 @@ public int getRelativeTop() {
     public String getPackageName() {
         return InstrumentationRegistry.getTargetContext().getPackageName();
     }
+
+    @Nullable
+    public String getViewTag() {
+        Object tag = view.getTag();
+        if (tag != null) {
```
mykola-mokhnach: return tag == null ? null : tag.toString()

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.java
```diff
@@ -179,6 +182,9 @@ public void describeTo(Description description) {
                     views = getViews(root, withXPath(selector), false);
                 }
                 break;
+            case VIEW_TAG:
+                views = getViews(root, withTagValue(allOf(instanceOf(String.class), equalTo((Object) selector))), findOne);
```
mykola-mokhnach: why findOne? Cannot it be that two views have the same tag?

PR #192, opened 2018-08-01
W3C /actions key Events

dpgraham:
* Wire up action events
* Maps unicode characters to Android KeyEvents
* Adjusted spec code so that it so that it accepts numbers as well as strings (which technically isn't allowed by the spec), because many of the Android Key Events don't have an obvious analagous unicode character
* Added e2e test and some unit tests


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -119,6 +168,158 @@ public void sychronousTickActionsComplete() throws AppiumException {
         multiTouchState.perform(uiController);
     }
 
+    public int getKeyCode(String keyValue, int location) throws AppiumException {
```
mykola-mokhnach: This implementation is ok, but I the supported chars range is pretty limited. We could use the same approach as adb though: https://android.googlesource.com/platform/frameworks/base.git/+/8f5e3d13e166e4873e130a5ef7c987333dbac917/cmds/input/src/com/android/commands/input/Input.java#148

dpgraham: Nice. Does that approach provide a full mapping between unicode and the key events, because ideally I would like to not have the number values.

mykola-mokhnach: I assume it provides the mapping between the currently selected IME. You could try it on your own by just executing `adb shell input text` with different values

dpgraham: I just tried the ADB way and it seems to work fine. This is a much better way.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -119,6 +278,123 @@ public void sychronousTickActionsComplete() throws AppiumException {
...
+                return KEYCODE_DPAD_DOWN;
+            case INSERT:
+                return KEYCODE_INSERT;
+            case DELETE:
+                return KEYCODE_FORWARD_DEL;
+        }
+
+        KeyCharacterMap keyCharacterMap = KeyCharacterMap.load(KeyCharacterMap.VIRTUAL_KEYBOARD);
+        android.view.KeyEvent[] keyEvent = keyCharacterMap.getEvents(keyValue.toCharArray());
+        return keyEvent[0].getKeyCode();
```
mykola-mokhnach: this is kind'a tricky. You must also consider syncing metastate, since many keys require also SHIFT to be pressed, for example uppercase stuff

dpgraham: I think it should be okay. The`globalKeyInput` keeps track of the meta state (if alt, ctrl or shift is being pressed). And I'm pretty sure the 'keyValue' should be case-insensitive, because I believe we're mapping the unicode character to a 'key' and not a 'character'.

mykola-mokhnach: you can try it on your own, but the actually generated key events sequence for character `H` will be:
Shift Down
Key h Down
Key h Up
Shift Up

Altogether 4 events and the first one does not contain the code of the character that we wanted

dpgraham: The shift unicode that I posted maps to the shift key according to the spec. I have a switch statement that checks if it's a shift and calls the Android shift keycode.


The test I have in there confirms that shift works as expected.

mykola-mokhnach: Perhaps I don't know something that you know.
Let say I execute

```java
new Actions(driver)
  .sendKeys("Hi")
  .perform()
```

this code will generate the following sequence according to https://github.com/SeleniumHQ/selenium/blob/d8f6fd34677b2662c0ac58e260efcc98f7ea3019/java/client/src/org/openqa/selenium/interactions/SendKeysAction.java

```
key down "H"
key up "H"
key down "i"
key up "i"
```

the current implementation on the server side will transform this into

```
key down SHIFT
key up SHIFT
key down "i"
key up "i"
```

dpgraham: The implementation, I would expect, would do:

```
key down "H"
key up "H"
key down "i"
key up "i"
```

but it would ignore the case.

If you want uppercase H, you need to to do keydown on 'SHIFT' before keydown on 'H'.

mykola-mokhnach: I assume users would expect the stuff on the client library side to simply work, shouldn't them? They don't really care about the internal implementation and nuances of the standard.

dpgraham:
Alright, I think it's good now @mykola-mokhnach....

Now, it pulls the KeyEvents array for the character that's being input and then dispatches each key event by bitwise-or'ing the global input state along with the key input state.

The test I have looks like this:

```javascript
[
      {"type": "keyDown", "value": "\uE008"},
      {"type": "keyDown", "value": "h"},
      {"type": "keyUp", "value": "h"},
      {"type": "keyDown", "value": "a"},
      {"type": "keyUp", "value": "a"},
      {"type": "pause", "duration": 2000},
      {"type": "keyUp", "value": "\uE008"},
      {"type": "keyDown", "value": "t"},
      {"type": "keyUp", "value": "t"},
      {"type": "keyDown", "value": "S"},
      {"type": "keyUp", "value": "S"},
    ];
```

And it will output HAtS

* H and A are capitalized because shift is down
* t stays lowercase
* S is uppercase because it's an uppercase character

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -1,56 +1,153 @@
...
+            final KeyEvent[] keyEventsFromChar = keyCharacterMap.getEvents(key.toCharArray());
+
+            if (keyEventsFromChar == null) {
+                throw new InvalidArgumentException(String.format("Could not find matching keycode for character %s", key));
+            }
+
+            int keyEventIndex = 0;
+            keyEvents = new ArrayList<>();
+
+
```
mykola-mokhnach: extra empty line

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -1,56 +1,153 @@
...
+        ));
+
+        // Save references to KeyEvents
+        if (isDown) {
+            keyDownEvents.put(key, keyEvents);
+        } else {
+            keyDownEvents.remove(key);
+        }
 
         try {
```
mykola-mokhnach: I assume this try/catch should only enclose injectKeyEvent method

PR #201, opened 2018-08-08
Add mouse click and doubleclick

dpgraham:


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/PointerEventHandler.java
```diff
@@ -38,8 +38,14 @@
     @Nullable
     private static MotionEvent globalTouchDownEvent;
     private static Map<Integer, MotionEvent> globalMouseButtonDownEvents = new HashMap<>();
+    private static Long globalMouseLocationX = 0L;
+    private static Long globalMouseLocationY = 0L;
     private static final DisplayMetrics displayMetrics = Resources.getSystem().getDisplayMetrics();
 
+    // Make the duration of a 'tap' or 'click' event half of the tap timeout
```
mykola-mokhnach: why should the duration divided by two?

dpgraham: Because the `ViewConfiguration.getTapTimeout()` and the `ViewConfiguration.getDoubleTapTimeout()` are the _maximum_ duration before it's not considered a tap/double-tap, thus I wanted the number to be less, so cutting it in half seemed good to me.

PR #204, opened 2018-08-09
Refactor press/long_press to use Espresso

dpgraham:
* Moved press/long_press into UIController performer
* Made it one handler instead of two with a parameter to specify if it's a longpress or not
* Use the LONG_PRESS_DURATION as the baseline for how long a press should be held for
* Also, unrelated to this PR, make Keys handler release key actions when done
  * The global meta state should be released after 'keys' is called
  * Also fixed the undoAll... don't call 'keyUp' if key is already released


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/Keys.java
```diff
@@ -67,11 +68,13 @@ public Void run(UiController uiController) throws AppiumException {
...
 
-                new ActionsBuilder()
+                Actions actions = new ActionsBuilder()
                         .withActions(Collections.singletonList(keyInputSource))
                         .withAdapter(new EspressoW3CActionAdapter(uiController))
-                        .build()
-                        .performActions(params.getSessionId());
+                        .build();
+
+                actions.performActions(params.getSessionId());
```
mykola-mokhnach: `Actions` suffix looks redundant for these methods

PR #206, opened 2018-08-14
Upgrade Node deps

dpgraham:
I published this driver yesterday and got the `helpers.installApkRemotely` is not defined problem, which was due to using appium-adb.... so I went ahead and just updated _all_ of Espresso Driver's dependencies.


mykola-mokhnach:
It would be interesting to know why greenkeeper does not take care of this stuff

dpgraham:
@mykola-mokhnach Yeah I'd like to know too. I don't think this repo has been getting the Greenkeeper PR's.

PR #210, opened 2018-08-16
Update tests to fix the build

imurchie:
Update the touch specs.
1. Use only a single session. This alone moves it from 6 minutes to ~50 seconds locally
2. Stop hardcoding sizes of things, so that the much smaller devices on Travis work.

### DIFF
--- a/test/functional/commands/touch-e2e-specs.js
+++ b/test/functional/commands/touch-e2e-specs.js
@@ -412,6 +412,19 @@ describe('touch actions -', function () {
 
         nextEl = await driver.elementByAccessibilityId('Next');
       });
+      it.only(`should perform single tap actions`, async function () { // eslint-disable-line
+        const {x, y} = await nextEl.getLocation();
+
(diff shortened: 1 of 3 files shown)


test/functional/commands/touch-e2e-specs.js
```diff
@@ -54,395 +97,416 @@ describe('elementByXPath', function () {
...
+            action.release();
+            action.perform();
+
+            await driver.elementByXPath("//*[@text='1']").should.eventually.exist;
+          });
+          it(`should perform single ${method} actions on an element`, async function () {
+            let action = new wd.TouchAction(driver);
+            action[method]({el: nextEl});
+            action.release();
+            action.perform();
```
mykola-mokhnach: should perform be always awaited?

imurchie: Yup. That was an existing problem. I thought I'd caught them all. :(

PR #212, opened 2018-08-21
Add an alias for ByTagName Selenium locator

mykola-mokhnach:
This will allow to use standard Selenium stuff without adding the extra stuff into client libs


saikrishna321:
@mykola-mokhnach in this case it will be `driver.findElementByTagName("")` on client side instead of `driver.findElementByAndroidViewTag("")` ?

mykola-mokhnach:
both will work with what is present in the pr

PR #214, opened 2018-08-22
Fix: Add 'pointerType' (touch, mouse, stylus) to inputState

dpgraham:
* input state didn't have a type before so it was maintaining the state by touch only
* This was causing wrong events to be dispatched
* Also
  * Fixed typo 'peformActions'
  * Removed test that checks that exception thrown if negative x,y actions... negative x, y is fine because it might be relative to a pointer or element

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/state/PointerInputState.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/state/PointerInputState.java
@@ -16,6 +16,10 @@ public class PointerInputState implements InputState {
     private long x = 0;
     private long y = 0;
 
+    public PointerInputState(PointerType pointerType) {
+        this.type = pointerType;
+    }
+
     public boolean isPressed(int num) {
         return pressed.contains(num);
     }
--- a/espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/PointerDispatchTest.java
+++ b/espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/PointerDispatchTest.java
@@ -35,6 +35,7 @@ import static io.appium.espressoserver.lib.helpers.w3c.dispatcher.PointerDispatc
 import static io.appium.espressoserver.lib.helpers.w3c.models.InputSource.ActionType.POINTER_MOVE;
 import static io.appium.espressoserver.lib.helpers.w3c.models.InputSource.ActionType.POINTER_UP;
 import static io.appium.espressoserver.lib.helpers.w3c.models.InputSource.POINTER;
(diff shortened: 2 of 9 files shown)


espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/PointerDispatchTest.java
```diff
@@ -67,7 +68,7 @@ public void shouldNoopPointerMoveIfNoButtons() throws ExecutionException, Interr
     public void shouldDoOneMoveIfDurationZero()
             throws ExecutionException, InterruptedException, AppiumException {
         DummyW3CActionAdapter dummyW3CActionAdapter = new DummyW3CActionAdapter();
-        pointerInputSource = new PointerInputState();
+        pointerInputSource = new PointerInputState(TOUCH);
```
mykola-mokhnach: this is not needed. Just add the default constructor, which assigns TOUCH type by default

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/state/PointerInputState.java
```diff
@@ -16,6 +16,10 @@
     private long x = 0;
     private long y = 0;
 
+    public PointerInputState(PointerType pointerType) {
+        this.type = pointerType;
```
mykola-mokhnach: type should be final, shouldn't it?

espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/ProcessorTest.java
```diff
@@ -122,30 +122,6 @@ public void shouldRejectPointerMoveIfNegativeDuration() throws InvalidArgumentEx
         }
     }
 
-    @Test
-    public void shouldRejectPointerMoveIfNegativeX() throws InvalidArgumentException {
-        Action action = new Action();
-        action.setX(-100);
-        try {
-            processPointerMoveAction(action, InputSourceType.POINTER, "any", 0);
```
mykola-mokhnach: I don't see any changes regarding this verification

dpgraham: It was a failing test from previous changed. It just wasn't caught because these tests don't run on CI. The issue is that I had made it fail for negative x y but it shouldn't automatically be doing that because you can have negative x y if it's a relative origin.

espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/PointerDispatchTest.java
dpgraham: I'd rather it be explicit because the reason this bug happened in the first place was because I didn't make it mandatory to set the touch type.

PR #215, opened 2018-08-22
Convert mouse to touch events if touchscreen device

dpgraham:
* Checks if the device has a touch screen, and if it does assume
mouse events are to be converted to touch
* This is because some clients only do mouse and we want those clients
to still be usable on touchscreen
* Also
  * Check downEvent is null before setting 'source' on pointer event. Pointer events don't always have
  * Improved logging for 'inject' event exception. Inject event always means it was outside the viewport

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/AndroidMotionEvent.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/AndroidMotionEvent.java
@@ -1,5 +1,6 @@
 package io.appium.espressoserver.lib.helpers.w3c.adapter.espresso;
 
+import android.support.annotation.Nullable;
 import android.support.test.espresso.UiController;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
@@ -32,7 +33,7 @@ public class AndroidMotionEvent {
                                     final Integer button,
                                     PointerType pointerType,
                                     final KeyInputState globalKeyInputState,
-                                    final MotionEvent downEvent,
+                                    @Nullable final MotionEvent downEvent,
                                     final long eventTime)
             throws AppiumException {
 
@@ -56,7 +57,7 @@ public class AndroidMotionEvent {
     public void pointerMove(List<Long> x, List<Long> y,
                             final PointerType pointerType,
                             final KeyInputState globalKeyInputState,
-                            final MotionEvent downEvent) throws AppiumException {
+                            @Nullable final MotionEvent downEvent) throws AppiumException {
         int metaState = getMetaState(globalKeyInputState);
 
         (new MotionEventBuilder())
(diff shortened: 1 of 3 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/AndroidMotionEvent.java
```diff
@@ -29,14 +29,14 @@ private AndroidMotionEvent(UiController uiController) {
 
     public MotionEvent pointerEvent(List<Long> x, List<Long> y,
                                     int action,
-                                    Integer button, PointerType pointerType,
+                                    final Integer button,
+                                    PointerType pointerType,
```
mykola-mokhnach: mark downEvent as Nullable

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -489,4 +493,25 @@ public void sleep(long duration) throws AppiumException {
     public Logger getLogger() {
         return AndroidLogger.logger;
     }
+
+    // If a 'mouse' event was provided, but it's a touchscreen, switch to 'touch'
+    // This is because some clients only send 'mouse' events and the assumption is that if they
+    // send 'mouse' events to a device that has a touch screen, it needs to be converted
+    // This may need to be re-visited in the future if we wish to support Android laptops
```
mykola-mokhnach: Android laptops usually have touch screen as well ;)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -489,4 +493,25 @@ public void sleep(long duration) throws AppiumException {
...
+    private boolean isTouch(PointerType type) {
+        return type == TOUCH || (type == MOUSE && isTouchScreen);
+    }
+
+    private boolean isTouchScreen() {
+        int[] deviceIds = InputDevice.getDeviceIds();
```
mykola-mokhnach: deviceIds intermediate variable seems to be redundant

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
dpgraham: I know. That's why we may need to revisit this in the future. It might be a problem for some to always have mouse actions converted to touch.

jlipps: which clients only support mouse? can we just send a simple PR to fix them instead?

mykola-mokhnach: it is not about they are only support mouse, but the default value is `mouse` there. And it is not so easy to change that

dpgraham:
PR fixes pushed

PR #216, opened 2018-08-22
Enhance motion event logging

dpgraham:


imurchie:
I'm all for better logging, but this seems to present a bunch of stuff but no information that is particularly useful. Could it be made readable by someone who is uninitiated?
```java
AndroidLogger.logger.info(String.format(
    "Running Android MotionEvent.obtain with parameters: \n" +
    "    downtime: [%s]\n" +
    "    event time: [%s]\n" +
    "    action: [%s]\n" +
    "    pointer count: [%s]\n" +
    "    pointer properties: [%s]\n" +
    "    pointer coordinates: [%s]\n" +
    "    meta state: [%s]\n" +
    "    button state: [%s]\n" +
    "    x precision: [%s]\n" +
    "    y precision: [%s]\n" +
    "    device id: [%s]\n" +
    "    edge flags: [%s]\n" +
    "    source: [%s]\n" +
    "    flags: [%s]" +
    "For more information, see https://developer.android.com/reference/android/view/MotionEvent#obtain(long,%20long,%20int,%20int,%20android.view.MotionEvent.PointerProperties[],%20android.view.MotionEvent.PointerCoords[],%20int,%20int,%20float,%20float,%20int,%20int,%20int,%20int),
    motionEventParams.downTime,
    eventTime,
    action,
    pointerCount,
    pointerProperties,
    pointerCoords,
    motionEventParams.metaState,
    motionEventParams.buttonState,
    motionEventParams.xPrecision,
    motionEventParams.yPrecision,
    motionEventParams.deviceId,
    motionEventParams.edgeFlags,
    motionEventParams.source,
    0
));
```

dpgraham:
@imurchie I think we may need to introduce different levels of logging to the Espresso driver.

I added this in because I wanted to be able to look at logcat crash logs and see that if this method caused an exception, what was passed into the method so I could understand why.

But I agree, for the normal Appium logs, this will get a bit unwieldy.

dpgraham:
Good call @imurchie, I updated the logging to that format.

PR #248, opened 2018-10-16
Fill out readme

dpgraham:


README.md
```diff
@@ -3,23 +3,32 @@
 [![Codacy Badge](https://api.codacy.com/project/badge/Grade/a877b7395f2d475aa79c08daf665dc3c)](https://www.codacy.com/app/dpgraham/appium-espresso-driver?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=appium/appium-espresso-driver&amp;utm_campaign=Badge_Grade)
 [![Greenkeeper badge](https://badges.greenkeeper.io/appium/appium-espresso-driver.svg)](https://greenkeeper.io/)
 
-Espresso integration for Appium
+Appium Espresso Driver is a test automation framework for Android that uses [Espresso](https://developer.android.com/training/testing/espresso/) as the underlying test technology. Appium Espresso Driver is part of the Appium framework.
```
mykola-mokhnach: perhaps `is a part`?

README.md
```diff
@@ -3,23 +3,32 @@
...
-```
-or
-```
-cd espresso-server
-./gradlew clean assembleDebug assembleAndroidTest
-```
+## Troubleshooting
 
-### Layout
+* If there are ever problems starting a session, try setting the capability `forceEspressoRebuild=true` and retrying. This will rebuild a fresh Espresso Server APK. If the session is succcesful, set it back to false so that it doesn't re-install on every single test.
```
mykola-mokhnach: We should also mention signing here. that is is required that both the server and the are signed with the same debug singature

dpgraham:
Fixes pushed. @mykola-mokhnach I also mentioned that there may be issues when using old version of SDK tools. I tried with an older version and it failed to sign. (But that was a _really_ old version and my machine is using Java 1)

PR #250, opened 2018-10-16
Check APK has internet permission

dpgraham:
* Add a check before session starts to verify that the permission is set
* Added a test that uses ContactManager.apk (which is known to not have that permission set) and write a test that confirms that it  fails to start sessions
* Updated package.json `npm run e2e-test` wasn't working... runs the raw mocha command


lib/driver.js
```diff
@@ -135,6 +135,13 @@ class EspressoDriver extends BaseDriver {
       if (this.opts.app) {
         // find and copy, or download and unzip an app url or path
         this.opts.app = await this.helpers.configureApp(this.opts.app, APP_EXTENSION);
+
+        // Check that it has the INTERNET permission set. This is required to use espresso-server.
+        const adb = await ADB.createADB();
```
mykola-mokhnach: it looks like a dirty hack. Can we instead move `this.adb` initialisation to a separate method instead of doing it inside `startEspressoSession`. Also, I cannot find where `checkPackagePresent` helper method is defined.

dpgraham: It's causing false session failures on CI as well.

I think what I'll do is just catch the error that occurs when the permission is _not_ set and then log a message + error recommending to set it.

mykola-mokhnach: yep, you can so it if you know which particular error message we need to handle

dpgraham:
Changes pushed. I'm just checking the output of the adb command that installs the APK for 'SocketException' and if it finds it, instruct to use INTERNET permission.

PR #252, opened 2018-10-17
Add mobile:swipe command

dpgraham:
Espresso has swipeUp, swipeDown, swipeLeft and swipeRight commands and they're mentioned in the cheatsheet so they must be important. Added them to the server.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/MobileSwipe.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/MobileSwipe.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.espressoserver.lib.handlers;
+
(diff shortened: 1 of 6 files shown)


lib/commands/general.js
```diff
@@ -6,6 +6,10 @@ commands.mobilePerformEditorAction = async function (opts = {}) {
   return await this.espresso.jwproxy.command('/appium/device/perform_editor_action', 'POST', {action});
 };
 
+commands.mobileSwipe = async function (opts = {}) {
+  const {direction, element} = opts;
```
mykola-mokhnach: should we unwrap the element first like we do in other drivers?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/MobileSwipe.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static android.support.test.espresso.action.ViewActions.swipeDown;
+import static android.support.test.espresso.action.ViewActions.swipeLeft;
+import static android.support.test.espresso.action.ViewActions.swipeRight;
+import static android.support.test.espresso.action.ViewActions.swipeUp;
+
+public class MobileSwipe implements RequestHandler<MobileSwipeParams, Void> {
+
+    @Override
+    @Nullable
```
mykola-mokhnach: it makes not much sense to mark Void method as nullable ;)

lib/commands/general.js
dpgraham: Is there an example of this. The only one I could find was in `appium-uiautomator2-driver`

```javascript
commands.mobileScrollBackTo = async function (opts) {
  const {elementId, elementToId} = opts;
  return await this.uiautomator2.jwproxy.command(`/appium/element/${elementId}/scroll_to/${elementToId}`, 'POST', {});
};
```

mykola-mokhnach: Very often I see that https://github.com/appium/appium-support/blob/db17f4f827c4e5e53b03481ae428cfcae25e9ff5/lib/util.js#L113 is used for such purpose. @imurchie Can you please provide more details on where this helper should be used?

saikrishna321:
https://github.com/appium/appium-espresso-driver/issues/259

PR #256, opened 2018-10-21
fix missing manifest

KazuCocoa:
Fix https://github.com/appium/appium/issues/11525#issuecomment-431633469

The cause is missing `espresso-server/AndroidManifest-test.xml` in npm package, it looks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
     "lib",
     "build/index.js",
     "build/lib",
-    "espresso-server/app/build/outputs/apk/androidTest/debug/app-debug-androidTest.apk"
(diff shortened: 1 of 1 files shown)


KazuCocoa:
Could you take a look and update the npm package again? @dpgraham

mykola-mokhnach:
There should be a way to prevent such issues in the future. We should either improve `copyFile`, so it throws a proper error if the source does not exist or add some verification into buildModServer method (less prefererable).

mykola-mokhnach:
@KazuCocoa Thank you for your investigation

KazuCocoa:
Agree 👍  https://github.com/appium/appium-espresso-driver/pull/256#issuecomment-431646541

Issue #258, opened 2018-10-23
1.9.2-beta.2 - Session termination not handled correctly


saikrishna321:
Logs : https://gist.github.com/saikrishna321/be7657ebe173033bd1b7f64db3ee5768

Comment from @sravanmedarapu

regarding bind failed issue.., it’s because of the server not being closed properly
https://github.com/appium/appium-espresso-driver/blob/master/espresso-server/app/src/androidTest/java/io/appium/espressoserver/EspressoServerRunnerTest.java#L43

will spend sometime over the weekend to implement session handling

and regarding intent.. looks like something going wrong here
https://github.com/appium/appium-espresso-driver/blob/7bd1fd13d6d2d3f6fe23a739057b009cebab7f1b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/StartActivity.java#L56

saikrishna321:
regarding wrong intent : https://gist.github.com/saikrishna321/262c2e66f305e9dc505a8793234f3b69#file-wrong-intent-handling-L147

Either App package and App activity of the previous session is picked

mykola-mokhnach:
I have an idea about what the problem can be

PR #265, opened 2018-10-24
Properly refresh the server package if the actual package is changed

mykola-mokhnach:
It might be that the target package is changed, so then we need to properly refresh the server apk. Unfortunately dumpsys says nothing about the target package, so we need to keep the target package value elsewhere to know when the server reinstall is needed

Addresses https://github.com/appium/appium-espresso-driver/issues/258


dpgraham:
When you get a chance, can you try rebasing master onto this? The Travis build appears to be working now.

mykola-mokhnach:
@dpgraham Yep, it's all good now. Thank you for the assistance.

PR #269, opened 2018-10-25
Speed up session startup

mykola-mokhnach:
Executing adb commands requires a lot of time. Getting the device information directly from the server is much more effective

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -1,6 +1,61 @@
 import _ from 'lodash';
+import { fs, tempDir } from 'appium-support';
+import log from '../logger';
 
 let commands = {}, helpers = {}, extensions = {};
+
+
+commands.getStrings = async function (language) {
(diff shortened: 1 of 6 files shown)


lib/commands/general.js
```diff
@@ -1,7 +1,62 @@
 import _ from 'lodash';
+import { fs, tempDir } from 'appium-support';
+import log from '../logger';
 
 let commands = {}, helpers = {}, extensions = {};
 
+
+commands.getStrings = async function (language) {
```
imurchie: Is there no way to not copy and paste this from `appium-uiautomator2-driver`?

mykola-mokhnach: I think this part of code might be changed in the future. also, it is not very clear where to put the common code

mykola-mokhnach: I think it would be an overkill

Issue #259, opened 2018-10-23
mobile:swipe sends element ID as undefined to proxy


saikrishna321:
Element Id is undefined when executing mobile commands 

Latest Appium beta:

https://gist.github.com/saikrishna321/ae546a5843569173179b388ab59bdfeb

```
MobileElement listview = (MobileElement) new WebDriverWait(driver, 30).
                until(ExpectedConditions.elementToBeClickable(MobileBy.AccessibilityId("listview")));
        Map<String, Object> args = new HashMap<>();
        args.put("element", listview.getId());
        args.put("direction", "up");
        driver.executeScript("mobile: swipe", args);
```

dpgraham:
@saikrishna321 Does `listview.getID()` return a UUID?

saikrishna321:
@dpgraham yes `d3daabe5-f8e3-44a5-9ac8-2bcbe29070e9`

```
[JSONWP Proxy] Replacing sessionId d4d34839-826b-4d52-af02-2cdcf93e2670 with b6fb78e5-efff-4b9c-8c79-b3d4bf425a2f
[HTTP] <-- GET /wd/hub/session/b6fb78e5-efff-4b9c-8c79-b3d4bf425a2f/element/d3daabe5-f8e3-44a5-9ac8-2bcbe29070e9/enabled 200 13 ms - 120
```

dpgraham:
This has been fixed

saikrishna321:
@dpgraham i have tested and it works with elementId and not element.

With element
```
debug] [JSONWP Proxy] Proxying [POST /appium/execute_mobile/[object Object]/swipe] to [POST http://localhost:8080/session/5ba78772-29e6-4222-9522-a4da3c54d81f/appium/execute_mobile/[object Object]/swipe] with body: {"direction":"up"}
[JSONWP Proxy] Got an unexpected response: {"id":"695ec959-c8e1-4474-913c-43ac1a112a4c","sessionId":null,"status":13,"value":"java.util.NoSuchElementException: Invalid element ID [object Object]\n\tat io.appium.espressoserver.lib.model.Element.getViewInteractionById(Element.java:56)\n\tat io.appium.espressoserver.lib.handlers.MobileSwipe....
[debug] [MJSONWP] Matched JSONWP error code 13 to UnknownError
```

With elementId
```
[...]

PR #270, opened 2018-10-26
respect element and elementId for mobile:swipe

saikrishna321:
XCUI-driver allows element and elementId. 

This PR allows espresso driver to accept both element and elementId to drivers consistent.

Related Issue: https://github.com/appium/appium-espresso-driver/issues/259

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import util from 'appium-support';
 
 let commands = {}, helpers = {}, extensions = {};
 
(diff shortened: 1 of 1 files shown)


lib/commands/general.js
```diff
@@ -22,7 +22,8 @@ commands.mobilePerformEditorAction = async function (opts = {}) {
 
 commands.mobileSwipe = async function (opts = {}) {
   const {direction, element} = assertRequiredOptions(opts, ['direction', 'element']);
-  return await this.espresso.jwproxy.command(`/appium/execute_mobile/${element}/swipe`, 'POST', {direction});
+  let elementId = element.ELEMENT || element;
```
mykola-mokhnach: please use `util.unwrapElement` from appium-support

mykola-mokhnach: also, this could be spelled like

`return await this.espresso.jwproxy.command(`/appium/execute_mobile/${util.unwrapElement(element)}/swipe`, 'POST', {direction})`

dpgraham: I put in a PR to make unwrapElement also unwrap W3C elements: https://github.com/appium/appium-support/pull/93/files

mykola-mokhnach: I'd also go through the codebase to make sure this helper is used everywhere instead of `manual` unwrapping

saikrishna321: @mykla-mokhnach should i wait for the unwrapElement which @dpgraham implemented to get merged ?

dpgraham: No need to wait. It will get the updated driver later.

PR #273, opened 2018-10-26
Fix tests, update moveTargetOutOfBounds logging

dpgraham:
* Fixed broken touch actions that fall out of the viewport for some devices/emulators
* Update move target out of bounds to include the viewport height and viewport width so we can troubleshoot how far out of bounds

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/MoveTargetOutOfBoundsException.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/MoveTargetOutOfBoundsException.java
@@ -2,14 +2,11 @@ package io.appium.espressoserver.lib.handlers.exceptions;
 
 
 public class MoveTargetOutOfBoundsException extends AppiumException {
-    public MoveTargetOutOfBoundsException() {
-        super("The target for pointer interaction is not in the viewport and cannot be brought into that viewport");
-    }
 
-    public MoveTargetOutOfBoundsException(long targetX, long targetY) {
+    public MoveTargetOutOfBoundsException(long targetX, long targetY, long viewportWidth, long viewportHeight) {
         super(String.format(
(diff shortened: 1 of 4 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/exceptions/MoveTargetOutOfBoundsException.java
```diff
@@ -2,14 +2,11 @@
 
 
 public class MoveTargetOutOfBoundsException extends AppiumException {
-    public MoveTargetOutOfBoundsException() {
-        super("The target for pointer interaction is not in the viewport and cannot be brought into that viewport");
-    }
 
-    public MoveTargetOutOfBoundsException(long targetX, long targetY) {
+    public MoveTargetOutOfBoundsException(long targetX, long targetY, long viewportWidth, long viewportHeight) {
```
mykola-mokhnach: I'd rather pass the Rect here and show its toShortString Description, since the actual bounds might not only be a screen bounds, but also element bounds

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/dispatcher/PointerDispatch.java
```diff
@@ -192,8 +192,10 @@ public static void dispatchPointerCancel(final W3CActionAdapter dispatcherAdapte
         }
 
         // 7-8. Bounds check
-        if (x < 0 || y < 0 || x > dispatcherAdapter.getViewportWidth() || y > dispatcherAdapter.getViewportHeight()) {
-            throw new MoveTargetOutOfBoundsException(x, y);
+        final long viewportWidth = dispatcherAdapter.getViewportWidth();
+        final long viewportHeight = dispatcherAdapter.getViewportHeight();
+        if (x < 0 || y < 0 || x > viewportWidth || y > viewportHeight) {
```
mykola-mokhnach: I think it would be more readable to use contains method of a Rect

Issue #276, opened 2018-10-31
Referencing elements changes even I did not re-call find_element after calling scroll


KazuCocoa:
When I tested below method, the results of `el.text` was changed even the `el` didn't change.

## Client (Ruby)
```ruby
)> el = @driver.find_element(:accessibility_id, 'Custom')
=> #<Selenium::WebDriver::Element:0x..face661ad4c2a5f0a id="dbafec13-1811-40f4-b592-468c8d077f55">
[23] pry(#<AppiumLibCoreTest::WebDriver::W3CActionsTest>)> el.text
=> "Custom"
# Scroll down and the _custom_ cell was gone like attached _After the 'Custom' gone_
[24] pry(#<AppiumLibCoreTest::WebDriver::W3CActionsTest>)> el.text
=> "Lists" # Wrong. I had not re-call find_element the _Lists_
```

`el.displayed?` also returns `true`. In this case, I beliave the `el.displayed?` should be false. `el.text` should return reference error or return _Custom_.

## Appium log
```
[W3C (61a0dbe9)] Driver proxy active, passing request on via HTTP proxy
[debug] [JSONWP Proxy] Matched '/wd/hub/session/61a0dbe9-7499-41c6-9162-098a81240ab2/element/dbafec13-1811-40f4-b592-468c8d077f55/text' to command name 'getText'
[debug] [JSONWP Proxy] Proxying [GET /wd/hub/session/61a0dbe9-7499-41c6-9162-098a81240ab2/element/dbafec13-1811-40f4-b592-468c8d077f55/text] to [GET http://localhost:8200/session/ed6ca2b0-9566-4ede-aa9f-19287910ad5e/element/dbafec13-1811-40f4-b592-468c8d077f55/text] with body: {}
[debug] [JSONWP Proxy] Got response with status 200: "{\"id\":\"14fbd1d7-1d6d-49f4-9024-29cc89dd3bfe\",\"sessionId\":\"ed6ca2b0-9566-4ede-aa9f-19287910ad5e\",\"status\":0,\"value\":\"Custom\"}"
[JSONWP Proxy] Replacing sessionId ed6ca2b0-9566-4ede-aa9f-19287910ad5e with 61a0dbe9-7499-41c6-9162-098a81240ab2
[HTTP] <-- GET /wd/hub/session/61a0dbe9-7499-41c6-9162-098a81240ab2/element/dbafec13-1811-40f4-b592-468c8d077f55/text 200 41 ms - 124
[...]

PR #278, opened 2018-10-31
Scroll elements back into view when scrolled out

dpgraham:
* Views that are part of an AdapterView don't always have the same data (`value`, `contentDescription`, etc...) throughout their lifecycle. This is because AdapterView stores data about elements in memory and only render views that are on-screen. When views get scrolled away from, instead of re-creating new views, views are re-cycled with updated `contentDescription`, `text`, etc... (see https://stackoverflow.com/questions/11945563/how-listviews-recycling-mechanism-works for reference)
* This will cause unexpected element caching results. If you locate a view, `el`, that is inside a ListView, and then the  ListView is scrolled, for instance, the Views will have their data "re-shuffled" and `el` will have different data.
* For the most part, we can't do anything about this because we cache Views when we locate elements, and there's no definitive test that confirms that a View was part of an AdapterView and had its data modified
* In one special condition, though, where this type of View has a "contentDescription", and the View was scrolled away from, when we retrieve the cached view, we can check if 'contentDescription' for the cached view (`actualContentDescription`) matches what it was when it was originally cached as (`expectedContentDescription`). If `actualContentDescription != expectedContentDescription`, then we can try scrolling `expectedContentDescription` into View, and updating the cache.

Other change in this  PR
* I removed the "loopUntilMainIdle" thread on retrieving elements, because that blocks the test in the case that we try to find an element by content description and plus it didn't fix the original problem of Views still being present

(related to this issue: https://github.com/appium/appium-espresso-driver/issues/276)

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
@@ -17,6 +17,7 @@
 package io.appium.espressoserver.lib.model;
 
 import android.support.test.espresso.DataInteraction;
+import android.support.test.espresso.EspressoException;
 import android.support.test.espresso.ViewInteraction;
 import android.view.View;
 import android.view.ViewParent;
@@ -26,6 +27,7 @@ import org.hamcrest.Matchers;
 
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -119,7 +121,7 @@ public class Element {
         if (contentDescriptionCache.containsKey(elementId)) {
             String expectedContentDesc = contentDescriptionCache.get(elementId);
 
-            if (view.getContentDescription() != null && !view.getContentDescription().equals(expectedContentDesc)) {
+            if (!Objects.equals(view.getContentDescription(), expectedContentDesc)) {
 
                 // Look up the view hierarchy to find the closest ancestor AdapterView
                 ViewParent ancestorAdapter = view.getParent();
@@ -141,7 +143,10 @@ public class Element {
                     view = (new ViewGetter()).getView(onView(withContentDescription(expectedContentDesc)));
                     cache.put(elementId, view);
                 } catch (Exception e) {
-                    throw new StaleElementException(elementId);
+                    if (e instanceof EspressoException) {
+                        throw new StaleElementException(elementId);
+                    }
+                    throw e;
                 }
             }
         }
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
@@ -16,9 +16,13 @@
 
 package io.appium.espressoserver.lib.model;
 
-import android.support.test.espresso.UiController;
+import android.support.test.espresso.DataInteraction;
 import android.support.test.espresso.ViewInteraction;
 import android.view.View;
+import android.view.ViewParent;
+import android.widget.AdapterView;
+
+import org.hamcrest.Matchers;
 
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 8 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
```diff
@@ -68,46 +85,67 @@ public static ViewInteraction getViewInteractionById(String elementId) throws Ap
...
+        //
+        // If the contentDescription of a cached view changed, it almost certainly means the rendering of an
+        // AdapterView (ListView, GridView, Scroll, etc...) changed and the View contents were shuffled around.
+        //
+        // If we encounter this, try to scroll the element with the expected contentDescription into
+        // view and return that element
+        if (contentDescriptionCache.containsKey(elementId)) {
+            String expectedContentDesc = contentDescriptionCache.get(elementId);
+
+            if (view.getContentDescription() != null && !view.getContentDescription().equals(expectedContentDesc)) {
```
mykola-mokhnach: can equals throw an NPE?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
```diff
@@ -68,46 +85,67 @@ public static ViewInteraction getViewInteractionById(String elementId) throws Ap
...
+                    );
+                    if (ancestorAdapter != null) {
+                        dataInteraction.inAdapterView(withView((AdapterView) ancestorAdapter));
+                    }
+                    dataInteraction.check(matches(isDisplayed()));
+
+                    // If successful, use that view instead of the cached view
+                    view = (new ViewGetter()).getView(onView(withContentDescription(expectedContentDesc)));
+                    cache.put(elementId, view);
+                } catch (Exception e) {
```
mykola-mokhnach: shall we log the original exception?

mykola-mokhnach: Perhaps it only makes sense to only filter Espresso exceptions instead of the generic one?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
mykola-mokhnach: !Objects.equals

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
```diff
@@ -68,46 +85,67 @@ public static ViewInteraction getViewInteractionById(String elementId) throws Ap
...
+
+                // Look up the view hierarchy to find the closest ancestor AdapterView
+                ViewParent ancestorAdapter = view.getParent();
+                while (ancestorAdapter != null && !(ancestorAdapter instanceof AdapterView)) {
+                    ancestorAdapter = ancestorAdapter.getParent();
+                }
 
-            if (!view.isShown()) {
-                throw new StaleElementException(elementId);
+                try {
```
mykola-mokhnach: ancestorAdapter might be null here

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
dpgraham: Does that work for String comparison though? The two strings being compared are different objects but I want to compare the values.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
dpgraham: I tried looking into narrowing down the Exceptions, but the exceptions are all runtime and I don't know which ones to catch.

mykola-mokhnach: I'd say https://developer.android.com/reference/android/support/test/espresso/EspressoException

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
```diff
@@ -68,46 +85,67 @@ public static ViewInteraction getViewInteractionById(String elementId) throws Ap
...
+                    // Try scrolling the view with the expected content description into the viewport
+                    DataInteraction dataInteraction = onData(
```
KazuCocoa: Ah... I see. Recycler View. 👍 > onData

dpgraham:
Fixes made

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/Element.java
```diff
@@ -68,46 +87,70 @@ public static ViewInteraction getViewInteractionById(String elementId) throws Ap
...
+                        dataInteraction.inAdapterView(withView((AdapterView) ancestorAdapter));
+                    }
+                    dataInteraction.check(matches(isDisplayed()));
+
+                    // If successful, use that view instead of the cached view
+                    view = (new ViewGetter()).getView(onView(withContentDescription(expectedContentDesc)));
+                    cache.put(elementId, view);
+                } catch (Exception e) {
+                    if (e instanceof EspressoException) {
+                        throw new StaleElementException(elementId);
```
mykola-mokhnach: it might still be useful to put the original exception as `cause`

PR #282, opened 2018-11-02
Fixed build problems

dpgraham:
* An update to dependencies broke unit tests, so did some refactoring. (I still haven't figured out how to run the Java unit tests from a command line, so if somebody wants to checkout this branch and try something, that would be greatly appreciated)
* Made some Codacy fixes


espresso-server/app/src/androidTest/java/io/appium/espressoserver/EspressoServerRunnerTest.java
```diff
@@ -34,22 +32,24 @@
...
     private final Server espressoServer = Server.getInstance();
 
     @Test
     public void startEspressoServer() throws InterruptedException, IOException, DuplicateRouteException {
-        try {
-            espressoServer.start();
-
-            while (!espressoServer.isStopRequestReceived()) {
-                Thread.sleep(1000);
+        System.out.println(System.getProperty("UNITTEST"));
```
mykola-mokhnach: This is not a good thing to put println into a production code

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/Rect.java
```diff
@@ -0,0 +1,43 @@
+package io.appium.espressoserver.lib.helpers;
+
+/**
+ * Copies parts of the Android Rect
+ *
+ * This is copied so that we can run in Java context
+ */
+public class Rect {
```
mykola-mokhnach: cannot we just use roboelectric instead?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/EspressoServerRunnerTest.java
dpgraham: That was  a mistake.

espresso-server/app/src/test/java/io/appium/espressoserver/test/assets/Helpers.java
```diff
@@ -8,7 +8,7 @@
 
     public static String readAssetFile(String filename) throws IOException {
```
mykola-mokhnach: Cannot we move these files to test resources instead?

PR #283, opened 2018-11-04
change default port from 8080 to 6791

saikrishna321:


lib/driver.js
```diff
@@ -120,7 +120,7 @@ class EspressoDriver extends BaseDriver {
         autoLaunch: true,
         adbPort: DEFAULT_ADB_PORT,
         androidInstallTimeout: 90000,
-        systemPort: 8080,
+        systemPort: 6791,
```
mykola-mokhnach: shouldn't it be set to DEVICE_PORT?

saikrishna321: @mykola-mokhnach yeah, my bad :) 

On second thought, do we really need to give _systemPort_ in defaultOpts ?

https://github.com/appium/appium-espresso-driver/blob/master/lib/driver.js#L144 this anyways finds free port and sets.

saikrishna321: FYI, UIA2 does not _systemPort_ in defaults

dpgraham: I think we should keep the default. It should be fine.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.java
```diff
@@ -37,7 +37,7 @@
 public class Server extends NanoHTTPD {
 
     private Router router;
-    private static final int DEFAULT_PORT = 8080;
+    private static final int DEFAULT_PORT = 6791;
```
dpgraham: It looks `systemPort` doesn't actually get used in this class. It's always set to `DEFAULT_PORT`.

lib/driver.js
dpgraham: We shouldn't have a default for `systemPort`. We should, by default, do what @saikrishna321 talked about and have it find a free port.

The port that's used on the actual device is a hardcoded port, 6791, which shouldn't cause any issues so as long as the Android device isn't also using port 6791 (which is extremely unlikely).

dpgraham: This is the block of code:

```this.opts.systemPort = this.opts.systemPort || await findAPortNotInUse(SYSTEM_PORT_RANGE[0], SYSTEM_PORT_RANGE[1]);```

If we set a default, it will never call `findAPortNotInUse`. So yes, we should remove the default.

saikrishna321:
@dpgraham as per comments i have removed _systemPort_ from default and let 'findAPortNotInUse' method allocate free port

PR #285, opened 2018-11-05
Support webviews

dpgraham:
Looks like we weren't supporting webviews on Espresso ¯ \_(ツ)_/¯

This PR ports over the exact same context-handling from `appium-uiautomator2-driver` and has webview tests that test switching to web context and running ChromeDriver commands and has a test for switching between webview _and_ native.

Also, we can't make use of the `browserName=Chrome` capability from Espresso because Chrome is a separate application and can't be accessed in an Espresso context. I updated it to throw an error if user tries using that capability.


lib/driver.js
```diff
@@ -124,6 +128,10 @@ class EspressoDriver extends BaseDriver {
       };
       _.defaults(this.opts, defaultOpts);
 
+      if (this.isChromeSession) {
+        logger.errorAndThrow(`Chrome browser cannot be run in Espresso sessions because Espresso automation doesn't have permission to access Chrome. Use UiAutomator2 instead`);
```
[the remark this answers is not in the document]
mykola-mokhnach: Also, I'd rather print a warning here, since grid might add this cap implicitly.

dpgraham:
@mykola-mokhnach Yeah I'm looking into it. It doesn't have Chromedriver.

lib/driver.js
```diff
@@ -123,6 +127,13 @@ class EspressoDriver extends BaseDriver {
       };
       _.defaults(this.opts, defaultOpts);
 
+      if (this.isChromeSession) {
+        logger.warn(`
+          'browserName' capability will be ignore.
```
mykola-mokhnach: ignored

mykola-mokhnach:
The changes are ok, but Travis is still not very happy

dpgraham:
@mykola-mokhnach Yeah I'm still sorting that out.

.travis.yml
```diff
@@ -65,6 +66,7 @@ install:
 
   # npm stuff
   - npm install
+  - npm install appium-chromedriver --chromedriver_version="2.6"
```
jlipps: why 2.6, isn't it very old? is it because there's a very old version of chrome there?

test/functional/webview/web-e2e-specs.js
```diff
@@ -0,0 +1,80 @@
...
+      await driver.title().should.eventually.equal('I am a page title');
+    });
+  });
+
+  describe('Chrome Browser', function () {
+    it('should reject "browserName=Chrome" sessions with a helpful error message (Chrome is not available in Espresso automation context)', async function () {
+      await initSession({
+        ...GENERIC_CAPS,
+        browserName: 'Chrome',
+      }).should.eventually.be.rejectedWith(/unavailable/);
```
jlipps: how will this pass? i don't see us actually throwing an error like this? also the test title is misleading and contains the text that should be in the assertion, not the title

.travis.yml
dpgraham: @jlipps Yeah, right now I'm just picking random versions of chromedriver to get working on Travis. Is there an easier way to determine which CD version is needed?

jlipps: I would say just get the latest, and then if it doesn't work you should see an error message which includes the version of chrome included on travis. Then you can find the highest version of chromedriver that supports that version of chrome.

jlipps: cf. https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/web/chromedriver.md

test/functional/webview/web-e2e-specs.js
dpgraham: When attempting to access an app that's not the AUT, Espresso throws an 'unavailable' error.

I was originally going to just throw an error if `browserName` was present, but @mykola-mokhnach wants to not fail sessions with `browserName` present just in case a client (like Selenium Grid) adds a `browserName` automatically.

dpgraham: I'll change the message though. TBH this test probably isn't even that important now.

jlipps: hmm, i would say we should fail at the soonest point we know something is wrong. just checking that browserName is present is probably not a good check. but if browserName is present and 'app' _isn't_, or if browserName is present and is equal to 'Chrome', well those are situations we know won't work, so i'd say we should throw at that point.

dpgraham:
I still can't get the tests running on Travis (they run fine locally). I made an issue for this: https://github.com/appium/appium-espresso-driver/issues/291

I'll get back to this soon, I just really need to get this release out.

PR #289, opened 2018-11-07
Use the native call to hide the keyboard

mykola-mokhnach:
Espresso has its own method to hide the keyboard and does not need to use adb


dpgraham:
I re-ran the tests and it worked. I'll have to look into this error: `Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds` some more.

PR #298, opened 2018-11-14
Switch xpath and source tree implementation to use StAX parser based on XmlPull

mykola-mokhnach:
This is the port of https://github.com/appium/appium-uiautomator2-server/pull/224


mykola-mokhnach:
@imurchie Do you have any comments to this PR?

PR #301, opened 2018-11-15
Fix server rebuild on application package change

mykola-mokhnach:
Addresses https://github.com/appium/appium-espresso-driver/issues/292#issuecomment-439026462

Now we store the package only after successful session init

### DIFF
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -46,21 +46,26 @@ class EspressoRunner {
   }
 
   async installTestApk () {
-    if (!this.forceEspressoRebuild && await this.isAppPackageChanged()) {
-      logger.info(`Forcing espresso server rebuild because of changed application package`);
-      this.forceEspressoRebuild = true;
+    let rebuild = this.forceEspressoRebuild;
+    if (rebuild) {
+      logger.debug(`'forceEspressoRebuild' capability is enabled`);
+    } else if (await this.isAppPackageChanged()) {
+      logger.info(`Forcing Espresso server rebuild because of changed application package`);
+      rebuild = true;
     }
 
-    if (this.forceEspressoRebuild && await fs.exists(this.modServerPath)) {
-      logger.debug(`Capability 'forceEspressoRebuild' on. Deleting file '${this.modServerPath}'`);
+    if (rebuild && await fs.exists(this.modServerPath)) {
+      logger.debug(`Deleting the obsolete Espresso server package '${this.modServerPath}'`);
       await fs.unlink(this.modServerPath);
     }
-
     if (!(await fs.exists(this.modServerPath))) {
       await this.buildNewModServer();
     }
-    await this.checkAndSignCert(this.modServerPath);
-    if (this.forceEspressoRebuild && await this.adb.uninstallApk(TEST_APK_PKG)) {
+    const isSigned = await this.adb.checkApkCert(this.modServerPath, TEST_APK_PKG);
+    if (!isSigned) {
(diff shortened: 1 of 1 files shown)


lib/espresso-runner.js
```diff
@@ -29,40 +29,28 @@ class EspressoRunner {
...
-      const dstPath = path.resolve(tmpRoot, path.posix.basename(TARGET_PACKAGE_CONTAINER));
-      await this.adb.pull(TARGET_PACKAGE_CONTAINER, dstPath);
-      const previousAppPackage = await fs.readFile(dstPath, 'utf8');
-      logger.debug(`The previous target application package was '${previousAppPackage}'. ` +
-        `The current package is '${this.appPackage}'.`);
-      return previousAppPackage !== this.appPackage;
-    } finally {
-      await fs.rimraf(tmpRoot);
-    }
+    const previousAppPackage = (await this.adb.shell(['cat', TARGET_PACKAGE_CONTAINER])).trim();
```
imurchie: This will just return something like `/system/bin/sh: cat: /data/local/tmp/espresso.fake.apppackage: No such file or directory` if `TARGET_PACKAGE_CONTAINER` doesn't exist, so the whole thing could be a single system call (to `adb shell cat <>`) and then simply a check for the string (if logging is something we care about, otherwise just the `return previousAppPackage !== this.appPackage;` line).

lib/espresso-runner.js
```diff
@@ -29,40 +29,28 @@ class EspressoRunner {
...
-    } finally {
-      await fs.rimraf(tmpRoot);
-    }
     logger.info(`Installed Espresso Test Server apk '${this.modServerPath}' (pkg: '${TEST_APK_PKG}')`);
   }
 
   async installTestApk () {
+    if (!this.forceEspressoRebuild && await this.isAppPackageChanged()) {
+      logger.info(`Forcing espresso server rebuild because of changed application package`);
+      this.forceEspressoRebuild = true;
```
imurchie: This seems odd. I don't think the object property should be changed when unnecessary. The fact that rebuild will have to happen is localized to this function, so a local variable would suffice.

imurchie: I think something like the following would be better
```js
let rebuild = this.forceEspressoRebuild;
if (rebuild) {
  logger.debug(`Capability 'forceEspressoRebuild' on.`);
} else if (await this.isAppPackageChanged()) {
  logger.info(`Forcing espresso server rebuild because of changed application package`);
  rebuild = true;
}
if (rebuild && await fs.exists(this.modServerPath)) {
  logger.debug(`Espresso server rebuilt required. Deleting file '${this.modServerPath}'`);
  await fs.unlink(this.modServerPath);
  await this.buildNewModServer();
}
await this.checkAndSignCert(this.modServerPath);
if (rebuild) {
  await this.adb.uninstallApk(TEST_APK_PKG);
  logger.info('Uninstalled the obsolete Espresso server package from the device under test');
}
```

lib/espresso-runner.js
mykola-mokhnach: this will make adb.shell to an exception, since cat'ing non-exisitng file returns exit code 1. Would you prefer exception-controlled over conditional-controlled flow?

lib/espresso-runner.js
mykola-mokhnach: makes sense, will change

lib/espresso-runner.js
imurchie: No, exception handling would be bad for this.

PR #303, opened 2018-11-19
Make "Actions" accept floats for "x", "y" and "duration" properties

dpgraham:
* Use floats as type for `x`, `y` and `duration` so that it doesn't break when a float is provided in Actions JSON
* When floats are provided in Actions JSON, round the values and provide warning to user if rounded value is different from provided value
* Also, add 'warn' as a log type (along with 'info', 'debug' and 'error')

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/AndroidLogger.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/AndroidLogger.java
@@ -47,5 +47,9 @@ public class AndroidLogger implements Logger {
         android.util.Log.d(TAG, toString(messages));
     }
 
+    public void warn(Object... messages) {
+        android.util.Log.w(TAG, toString(messages));
+    }
+
     public static Logger logger = new AndroidLogger();
 }
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/DummyW3CActionAdapter.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/DummyW3CActionAdapter.java
@@ -24,10 +24,10 @@ public class DummyW3CActionAdapter extends BaseW3CActionAdapter {
     public static class PointerMoveEvent {
         public String sourceId;
         public PointerType pointerType;
-        public long currentX;
-        public long currentY;
-        public long x;
-        public long y;
+        public float currentX;
+        public float currentY;
+        public float x;
+        public float y;
         public Set<Integer> buttons;
         public KeyInputState globalKeyInputState;
     }
@@ -45,6 +45,9 @@ public class DummyW3CActionAdapter extends BaseW3CActionAdapter {
         public void debug(Object... messages) {
             // No-op
         }
+        public void warn(Object... messages) {
+            // No-op
+        }
     }
 
     public void keyDown(W3CKeyEvent keyEvent) {
@@ -56,18 +59,18 @@ public class DummyW3CActionAdapter extends BaseW3CActionAdapter {
     }
 
     public void pointerUp(int button, String sourceId, PointerType pointerType,
-                            Long x, Long y, Set<Integer> depressedButtons,
+                            Float x, Float y, Set<Integer> depressedButtons,
                             KeyInputState globalKeyInputState) throws AppiumException {
         // No-op
     }
 
     public void pointerDown(int button, String sourceId, PointerType pointerType,
-                     Long x, Long y, Set<Integer> depressedButtons,
+                            Float x, Float y, Set<Integer> depressedButtons,
                      KeyInputState globalKeyInputState) throws AppiumException {
         // No-op
     }
 
-    public void pointerCancel(String sourceId, PointerType pointerType) throws AppiumException {
+    public void pointerCancel(String sourceId, PointerType pointerType) {
         // No-op
     }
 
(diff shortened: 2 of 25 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/AndroidLogger.java
```diff
@@ -47,5 +47,9 @@ public void debug(Object... messages) {
         android.util.Log.d(TAG, toString(messages));
     }
 
+    public void warn(Object... messages) {
+        android.util.Log.w(TAG, toString(messages));
```
jlipps: will this ultimately show up in the driver logs? that's probably where we will want it...

dpgraham: @jlipps yeah, these are the logs that show up in the Appium logs.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/DummyW3CActionAdapter.java
```diff
@@ -85,7 +88,7 @@ public double getPointerMoveDurationMargin(PointerInputState pointerInputState)
     }
 
     public void pointerMove(String sourceId, PointerType pointerType,
-                            long currentX, long currentY, long x, long y,
+                            float currentX, float currentY, float x, float y,
```
mykola-mokhnach: This is not very consistent. Sometimes `float` is used and in other places `Float`. Also FYI: JSON parser transforms non-integer numbers to Double by default, so we do perform an unnecessary type conversion here

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -480,8 +483,12 @@ public void waitForUiThread() {
         uiController.loopMainThreadUntilIdle();
     }
 
-    public void sleep(long duration) throws AppiumException {
-        uiController.loopMainThreadForAtLeast(duration);
+    public void sleep(float duration) {
+        long roundedDuration = Math.round(duration);
+        if (duration != roundedDuration) {
```
mykola-mokhnach: `float != int` should always be true. One must cast float to an integer type first in order to compare it for equality

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
```diff
@@ -495,4 +502,23 @@ private boolean isTouch(PointerType type) {
...
+     * Gives warning if the long values are different from the float values
+     *
+     * @param x X coordinate
+     * @param y Y coordinate
+     * @return Rounded x and y coordinates
+     */
+    private Point getRoundedCoordinates(float x, float y) {
+        int roundedX = Math.round(x);
+        int roundedY = Math.round(y);
+        if (x != roundedX || y != roundedY) {
```
mykola-mokhnach: see above

espresso-server/app/src/test/java/io/appium/espressoserver/test/helpers/w3c/ActionSequenceTest.java
```diff
@@ -48,17 +48,17 @@ public void shouldPullOutPointerActionsInW3CActions() throws IOException, Invali
         assertEquals(action.getId(), "finger1");
         assertEquals(action.getType(), POINTER);
         assertEquals(action.getSubType(), POINTER_MOVE);
-        assertEquals(action.getDuration(), new Long(0));
-        assertEquals(action.getX(), 100);
-        assertEquals(action.getY(), 100);
+        assertEquals(action.getDuration(), new Float(0));
+        assertEquals(action.getX(), 100, 1e-15);
```
mykola-mokhnach: you can use `Math.ulp(1.0)` instead of the hardcoded number

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/DummyW3CActionAdapter.java
dpgraham: The usage of primitives vs. objects is admittedly arbitrary. When I was working on the Actions api I often made the choice based on whichever passed

mykola-mokhnach: I would just prefer the stuff to be consistent everywhere

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/adapter/espresso/EspressoW3CActionAdapter.java
dpgraham: I just did this test:

```java

        long numOne = 10;
        float numTwo = 10.0F;
        boolean comp = numOne == numTwo;
```

And `comp` evaluates to true.

I think that when comparisons are made, `long` is cast to `float`.

PR #316, opened 2018-11-26
Add mobile: endpoint to verify toast messages presence

mykola-mokhnach:
`mobile: isToastVisible`

Arguments:
`text` - mandatory argument. Either exact toast text or a regular expression (depending on `isRegexp ` value)
`isRegexp` - optional argument. If set to `true` then `text` will be parsed as regular expression pattern. By default the matching by exact toast text is performed.


mykola-mokhnach:
@KazuCocoa FYI

PR #319, opened 2018-11-27
Implement 'requestPermissions'

dpgraham:
(see https://developer.android.com/reference/androidx/test/runner/permission/PermissionRequester)

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/RequestPermissionsHandler.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/RequestPermissionsHandler.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.espressoserver.lib.handlers;
+
+import java.util.regex.Pattern;
+
+import androidx.test.espresso.EspressoException;
+import androidx.test.espresso.ViewInteraction;
+import androidx.test.runner.permission.PermissionRequester;
+import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.model.RequestPermissionsParams;
+import io.appium.espressoserver.lib.model.ToastLookupParams;
+import io.appium.espressoserver.lib.viewmatcher.ToastMatcher;
+
+import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.assertion.ViewAssertions.matches;
+import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static androidx.test.espresso.matcher.ViewMatchers.withText;
+import static io.appium.espressoserver.lib.viewmatcher.RegexpTextMatcher.withRegexp;
+
+public class RequestPermissionsHandler implements RequestHandler<RequestPermissionsParams, Void> {
+    @Override
+    public Void handle(RequestPermissionsParams params) throws AppiumException {
+        final PermissionRequester permissionRequester = new PermissionRequester();
+        for (final String permission: params.getPermissions()) {
+            try {
+                permissionRequester.addPermissions(permission);
+            } catch (Throwable e) {
+                throw new AppiumException(String.format("Could not grant permission '%s'. Reason: %s", permission, e.getMessage()));
+            }
+        }
+        try {
+            permissionRequester.requestPermissions();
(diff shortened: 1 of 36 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/ScreenshotHandler.java
```diff
@@ -16,15 +16,25 @@
...
+
+import androidx.test.runner.screenshot.ScreenCapture;
+import androidx.test.runner.screenshot.Screenshot;
 import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
-import io.appium.espressoserver.lib.helpers.ScreenshotsHelper;
 import io.appium.espressoserver.lib.model.AppiumParams;
-import io.appium.espressoserver.lib.viewaction.ViewGetter;
 
-public class Screenshot implements RequestHandler<AppiumParams, String> {
+public class ScreenshotHandler implements RequestHandler<AppiumParams, String> {
```
jlipps: what does this change have to do with permissions?

lib/commands/general.js
```diff
@@ -42,6 +42,18 @@ commands.mobileIsToastVisible = async function (opts = {}) {
   });
 };
 
+commands.mobileRequestPermissions = async function (opts = {}) {
```
jlipps: why make this a mobile: method? i'd prefer to use the `permissions` cap we're already using for ios. or if we want permissions to be granted mid-run, use the appium settings API.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/RequestPermissionsHandler.java
```diff
@@ -0,0 +1,41 @@
...
+import io.appium.espressoserver.lib.model.RequestPermissionsParams;
+
+public class RequestPermissionsHandler implements RequestHandler<RequestPermissionsParams, Void> {
+    @Override
+    public Void handle(RequestPermissionsParams params) throws AppiumException {
+        final PermissionRequester permissionRequester = new PermissionRequester();
+        for (final String permission: params.getPermissions()) {
+            try {
+                permissionRequester.addPermissions(permission);
+            } catch (Throwable e) {
```
mykola-mokhnach: It is not recommended to catch Throwable. Replace it with Exception

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/RequestPermissionsHandler.java
```diff
@@ -0,0 +1,41 @@
...
+                throw new AppiumException(String.format("Could not grant permission '%s'. Reason: %s", permission, e.getMessage()));
```
mykola-mokhnach: I'd add the original exception as `cause`

lib/commands/general.js
```diff
@@ -42,6 +42,18 @@ commands.mobileIsToastVisible = async function (opts = {}) {
...
+  let {permissions} = opts;
+  if (!util.hasValue(permissions)) {
+
```
mykola-mokhnach: why it is empty?

lib/commands/general.js
mykola-mokhnach: It probably makes sense to grant them in runtime as well, but the method should be more universal, so we can do both - grant and revoke a particular permission.

dpgraham: I couldn't find an API method that revokes permissions.

mykola-mokhnach: we can still do it via adb. androidx also contains helpers to execute shell commands

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/RequestPermissionsHandler.java
```diff
@@ -0,0 +1,41 @@
...
+import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.model.RequestPermissionsParams;
+
+public class RequestPermissionsHandler implements RequestHandler<RequestPermissionsParams, Void> {
+    @Override
+    public Void handle(RequestPermissionsParams params) throws AppiumException {
+        final PermissionRequester permissionRequester = new PermissionRequester();
+        for (final String permission: params.getPermissions()) {
+            try {
+                permissionRequester.addPermissions(permission);
```
KazuCocoa: What about defending  `params.getPermissions()` as `String[]`?
Then, we can get rid of `for` and can call ` permissionRequester.addPermissions(params.getPermissions());`, simply.

dpgraham: `addPermissions` accepts Strings... I don't think it accepts `String[]`

KazuCocoa: ah, I have simply implemented the same code as:

```kotlin
    private fun grantPermissions(permissions: Array<String>) {
        val requester = PermissionRequester()
        requester.addPermissions(*permissions)
        requester.requestPermissions()
    }
```

The converted to Java code was:

```java
   private final void grantPermissions(String[] permissions) {
      PermissionRequester requester = new PermissionRequester();
      requester.addPermissions((String[])Arrays.copyOf(permissions, permissions.length));
      requester.requestPermissions();
   }
```

So I thought so.

dpgraham:
I'm going to close this for now. I won't have time to prioritize this issue, and by the sounds of it we already have this functionality through ADB.

Issue #320, opened 2018-11-27
Refactor SourceDocument to use TreeIterables


dpgraham:
AndroidX has helper methods that traverse the view hierarchy (https://developer.android.com/reference/androidx/test/espresso/util/TreeIterables)

We should be using this to construct the SourceDocument

PR #321, opened 2018-11-27
Use androidx source tree traverser

mykola-mokhnach:
Addresses https://github.com/appium/appium-espresso-driver/issues/320


dpgraham:
@mykola-mokhnach Yep.... if there's a problem with this, the source specs and the XML finding specs should catch it.

PR #322, opened 2018-11-27
Use GeneralLocation library to get center point

dpgraham:
Espresso has a library that calculates the location of a View. Might as well use that instead of our own calculations.


mykola-mokhnach:
@dpgraham Please rebase the PR with master to pass the CI

PR #325, opened 2018-11-28
Fix NPE in index calculation

mykola-mokhnach:
Addresses https://travis-ci.org/appium/appium-espresso-driver/jobs/460491765#L6095


mykola-mokhnach:
Linking https://github.com/appium/appium-espresso-driver/pull/322

PR #330, opened 2018-12-01
Update deprecated routing methods

mykola-mokhnach:
Also adds proper error description in case the application under test does not declare internet permission in its manifest.


mykola-mokhnach:
Linking https://gist.github.com/saikrishna321/075362c4e42b582a3bd87af5005b8cd2

PR #335, opened 2018-12-04
Add feature to invoke method on element

rajdeepv:
This is yet another feature which I have used a lot in Calabash. Using this feature we can do a lot with element **without modifying App under test**.

**Examples** 

1) Sometimes element visible property is true, but its not actually visible because of [alpha](https://developer.android.com/reference/android/view/View.html#ALPHA) is 0
In this case we can get alpha as:
`@driver.execute_script("mobile: elementInvoke", {element:id, methods:[{name:'getAlpha'}]})`

2) If I want to get progress of a progressbar or seekbar element:
`@driver.execute_script("mobile: elementInvoke", {element:<element_id>, methods:[{name:'getProgress'}]})
`
3) If I want to check if a text is _Italic_
`@driver.execute_script("mobile: elementInvoke", {element:<element_id>, methods:[{name:'getTypeface'}, {name:'isItalic'}]})
`

These are simple examples, but I have some really complex tasks which otherwise is not possible to do. 

It doesn't need modification of app under test and works with ReactNative. It reuses backdoor code


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/ElementInvoke.java
```diff
@@ -0,0 +1,18 @@
...
+import io.appium.espressoserver.lib.handlers.exceptions.AppiumException;
+import io.appium.espressoserver.lib.helpers.BackdoorUtils;
+import io.appium.espressoserver.lib.helpers.InvocationOperation;
+import io.appium.espressoserver.lib.model.Element;
+import io.appium.espressoserver.lib.model.ElementInvokeParams;
+
+public class ElementInvoke implements RequestHandler<ElementInvokeParams, Object> {
+    @Override
+    public Object handle(ElementInvokeParams params) throws AppiumException {
+        Object view = Element.getViewById(params.getElementId());
```
mykola-mokhnach: getViewById returns View

rajdeepv: changed

lib/commands/general.js
```diff
@@ -176,6 +176,11 @@ commands.mobileBackdoor = async function (opts = {}) {
   return await this.espresso.jwproxy.command(`/appium/execute_mobile/backdoor`, 'POST', {target, methods});
 };
 
+commands.mobileElementInvoke = async function (opts = {}) {
```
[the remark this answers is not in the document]
rajdeepv: added

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/BackdoorUtils.java
```diff
@@ -46,4 +54,32 @@ public static Object parseValue(Object o, Class<?> c) {
...
+        return invocationResult;
+    }
+
+    public static List<InvocationOperation> getOperations(List<MobileBackdoorMethod> mobileBackdoorMethods) throws InvalidArgumentException {
+        List<InvocationOperation> ops = new ArrayList<>();
+
+        for (MobileBackdoorMethod mobileBackdoorMethod : mobileBackdoorMethods) {
+            String methodName = mobileBackdoorMethod.getName();
+            if (methodName == null) {
+                throw new InvalidArgumentException("'name' is a required parameter for backdoor method to be invoked.");
```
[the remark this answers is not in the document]
rajdeepv: changed

lib/commands/general.js
```diff
@@ -176,6 +176,52 @@ commands.mobileBackdoor = async function (opts = {}) {
...
+ *
+ *
+ * @throws  {Error} if a method is not found with given argument types
+ *
+ * @return {*} the result of the last method in the invocation chain. If method return type is void, then "<VOID>" will be returned
+ *
+ */
+commands.mobileElementInvoke = async function (opts = {}) {
+  const {element, methods} = assertRequiredOptions(opts, ['element', 'methods']);
+  return await this.espresso.jwproxy.command(`/appium/execute_mobile/${util.unwrapElement(element)}/invoke_method`, 'POST', {methods});
```
mykola-mokhnach: do we really need a separate endpoint for this? I think we could instead simply add one more target (aka element id) type into `mobile: backdoor` call

rajdeepv: Usually, I see elementId is a uri parameter, therefore tried to follow same practice and created a different endpoint. I can add another target "element" and additional parameter elementId in the POST

mykola-mokhnach:
Less code -> less bugs ;)

PR #343, opened 2018-12-11
Test source document

dpgraham:
* Add a test that would cause a Java heap error, if not for the XML traversal depth limit introduced by: https://github.com/appium/appium-espresso-driver/pull/341
* Runs the tests on Travis CI now. Skips the Espresso Server test by adding a property that tells the test to be skipped.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
@@ -67,7 +67,7 @@ public class SourceDocument {
     private static final String VIEW_INDEX = "viewIndex";
     private static final String NAMESPACE = "";
     private static final String DEFAULT_VIEW_CLASS_NAME = "android.view.View";
-    private static final int MAX_TRAVERSAL_DEPTH = 70;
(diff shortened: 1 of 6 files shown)


mykola-mokhnach:
Looks good, but hold on with tests please, until the original problem is fully addressed.

dpgraham:
@mykola-mokhnach How about we commit the tests now, and then afterwards I'll update the tests to reflect any new changes.

mykola-mokhnach:
I would prefer to wait. But if you are in rush...

dpgraham:
@mykola-mokhnach Sure... do you have any recommendations for other cases that we can test on SourceDocument? I'd like to get pretty good coverage.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.java
```diff
@@ -317,4 +317,8 @@ public String toXMLString() throws AppiumException {
             throw new XPathLookupException(xpathSelector, xe.getMessage());
         }
     }
+
+    public static void $setMaxTraverseDepth(int maxTraverseDepth) {
```
mykola-mokhnach: I'd rather mark it as visibleForTesting

PR #357, opened 2018-12-17
uninstall espresso server if the installed server version is newer

KazuCocoa:
https://github.com/appium/appium-uiautomator2-driver/pull/270 for Espresso.

`installOrUpgrade` includes the same, lower version and unknown cases. The only newer version does not exist in the method.

I considered implementing `uninstallApk` when installed app is `NEWER_VERSION_INSTALLED` in `installOrUpgrade`. But current behaviour is proper for the current method name. The cost of calling `getApplicationInstallState` is not considerable, I thought.

Thus, I add this way.

### DIFF
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -41,6 +41,12 @@ class EspressoRunner {
   }
 
   async installServer () {
+    // this.adb.installOrUpgrade remains newer Espresso server
+    const appState = await this.adb.getApplicationInstallState(this.modServerPath, TEST_APK_PKG);
+    if (appState === this.adb.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED) {
+      await this.adb.uninstallApk(TEST_APK_PKG);
+    }
+
     await this.adb.installOrUpgrade(this.modServerPath, TEST_APK_PKG);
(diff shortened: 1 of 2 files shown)


lib/espresso-runner.js
```diff
@@ -41,6 +41,12 @@ class EspressoRunner {
   }
 
   async installServer () {
+    // this.adb.installOrUpgrade remains newer Espresso server
+    const appState = await this.adb.getApplicationInstallState(this.modServerPath, TEST_APK_PKG);
+    if (appState === this.adb.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED) {
+      await this.adb.uninstallApk(TEST_APK_PKG);
+    }
+
     await this.adb.installOrUpgrade(this.modServerPath, TEST_APK_PKG);
```
mykola-mokhnach: regarding the cost of getApplicationInstallState call - it is called inside installOrUpgrade for the second time. This means we should consider changing `installOrUpgrade` to `install`, like UIA2 driver does if the upgrade logic needs to be customized

KazuCocoa: yeah..

KazuCocoa:
https://github.com/appium/appium-espresso-driver/pull/358

PR #360, opened 2018-12-20
Throw correct exception if a screenshot cannot be taken

mykola-mokhnach:


mykola-mokhnach:
No, adb screenshot will most likely fail as well if this API fails.

PR #369, opened 2019-01-11
add settings to return empty results

KazuCocoa:
settings apis return no route error.

Espresso driver has no settings api, since it has no settings.
I wonder 
- we add API, but it returns empty or
- clarifies Espresso driver does not support settings API in https://github.com/appium/appium/tree/16f545c37d730fca431fcd2cb9767221312738dd/commands-yml/commands/session/settings


KazuCocoa:
Switched with not implemented yet error.
Does this make sense than the previous one?

lib/commands/general.js
```diff
@@ -197,6 +198,16 @@ commands.flashElement = async function (opts = {}) {
   });
 };
 
+// eslint-disable-next-line require-await,no-unused-vars
+commands.updateSettings = async function (_settings) {
```
mykola-mokhnach: the underscore in the var name is not needed

PR #377, opened 2019-01-21
Add `installTimeout` and add Kotlin

dpgraham:
* `androidInstallTimeout` was being ignored in `appium-espresso-driver`. Update it to include `androidInstallTimeout` as a paramter in `this.adb.install` and as a member variable in `EspressoRunner
* Update gradle to use Kotlin

### DIFF
--- a/espresso-server/app/build.gradle
+++ b/espresso-server/app/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
 
 android {
     compileSdkVersion 28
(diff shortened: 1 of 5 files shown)


espresso-server/app/build.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
```
mykola-mokhnach: wasn't this added in the other PR?

espresso-server/build.gradle
```diff
@@ -1,6 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.kotlin_version = '1.3.11'
```
mykola-mokhnach: ..and this

espresso-server/app/build.gradle
dpgraham: Yeah, I'm going to fix that up.... I wanted to narrow down the problem, so I made a PR that only has the gradle changes.

PR #385, opened 2019-01-23
[fix the build] Add clickAction to mobile mapping

imurchie:
`clickAction` was missing from the mobile command mapping.


dpgraham:
Thanks @imurchie , It was mistakenly removed during a rebase.

PR #398, opened 2019-02-13
Refactor elements cache implementation

mykola-mokhnach:
This is the second try to revamp  the current stuff and to make it nice and easy to read/modify


mykola-mokhnach:
I don't think there is much to test. The Cache class is just a thin wrapper over a hash map and the other stuff is already covered by integration tests

PR #415, opened 2019-03-14
Add permissions handling mobile endpoints

mykola-mokhnach:
Based on https://github.com/appium/appium-android-driver/pull/514


dpgraham:
This is a good example of why I think we should use Lerna and have all our packages in one repo. These could have all been done in one pr instead of three.

PR #446, opened 2019-05-06
Update qualifyActivityName algorithm

mykola-mokhnach:
Borrowed from https://android.googlesource.com/platform/frameworks/base/+/master/tools/aapt/Resource.cpp#755


dpgraham:
I have another PR in that will fix the webview error so I'll just merge it as is.

Issue #453, opened 2019-05-22
Backdoor on target element fails: Cannot find 'null' element


TomBAMU:
I'm using the current appium beta version. 

I tried to retrieve the text color of a textview by using "mobile:backdoor" script with the espresso driver. Unfortunately I am running into a InvalidArgumentException.

I'm not sure why, since the arguments contain an elementId.

Code to reproduce
```
Map<String, Object> args = new HashMap<String, Object>() {{
            put("target", "element");
            put("elementId", element.getId());
            put("methods", Collections.singletonList(ImmutableMap.of(
                    "name", "hasTextColor"
            )));
        }};
androidDriver.executeScript("mobile:backdoor", args);
```

The server log:
```
debug] [WD Proxy] Matched '/wd/hub/session/f3b2c06c-ea39-49d6-a6a0-8fd4ed985c4a/element' to command name 'findElement'
[debug] [WD Proxy] Proxying [POST /wd/hub/session/f3b2c06c-ea39-49d6-a6a0-8fd4ed985c4a/element] to [POST http://localhost:8012/session/1203e6be-db68-4b6d-95f5-efa5bb869c49/element] with body: {"using":"xpath","value":"//*[@text='74162103620129']"}
[debug] [WD Proxy] Got response with status 200: {"id":"d0da4986-163c-4e55-ad75-f46b5ae9a47e","sessionId":"1203e6be-db68-4b6d-95f5-efa5bb869c49","status":0,"value":{"ELEMENT":"03dfa766-a73d-44bd-81a6-d81c809a4791"}}
[WD Proxy] Replacing sessionId 1203e6be-db68-4b6d-95f5-efa5bb869c49 with f3b2c06c-ea39-49d6-a6a0-8fd4ed985c4a
[HTTP] <-- POST /wd/hub/session/f3b2c06c-ea39-49d6-a6a0-8fd4ed985c4a/element 200 740 ms - 166
[HTTP]
[HTTP] --> GET /wd/hub/session/f3b2c06c-ea39-49d6-a6a0-8fd4ed985c4a/element/03dfa766-a73d-44bd-81a6-d81c809a4791/location
[HTTP] {}
[W3C (f3b2c06c)] Driver proxy active, passing request on via HTTP proxy
[...]

PR #405, opened 2019-02-21
Refactor models to Kotlin (A ~> M)

dpgraham:

PR #454, opened 2019-05-24
fix element id in mobile backdoor

KazuCocoa:
closes https://github.com/appium/appium-espresso-driver/issues/453

`targetElement`  is used by Gson in MobileBackdoorParams to map a parameter instead of `elementId` after https://github.com/appium/appium-espresso-driver/pull/405

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -204,8 +204,8 @@ commands.mobileBackdoor = async function mobileBackdoor (opts = {}) {
   if (target === 'element') {
     assertRequiredOptions(opts, ['elementId']);
   }
-  const {elementId} = opts;
-  return await this.espresso.jwproxy.command(`/appium/execute_mobile/backdoor`, 'POST', {target, methods, elementId});
+  const {elementId: targetElement} = opts;
+  return await this.espresso.jwproxy.command(`/appium/execute_mobile/backdoor`, 'POST', {target, methods, targetElement});
 };
 
 /**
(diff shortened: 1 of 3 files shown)


rajdeepv:
Thanks @KazuCocoa

KazuCocoa:
(feel free to merge this since i have no merge permission)

imurchie:
@KazuCocoa That should not be the case, now.

lib/commands/general.js
```diff
@@ -204,8 +204,8 @@ commands.mobileBackdoor = async function mobileBackdoor (opts = {}) {
   if (target === 'element') {
     assertRequiredOptions(opts, ['elementId']);
   }
-  const {elementId} = opts;
-  return await this.espresso.jwproxy.command(`/appium/execute_mobile/backdoor`, 'POST', {target, methods, elementId});
+  const {elementId: targetElement} = opts;
+  return await this.espresso.jwproxy.command(`/appium/execute_mobile/backdoor`, 'POST', {target, methods, targetElement});
```
mykola-mokhnach: wouldn't it be easier to fix it in the server code instead?

KazuCocoa: No, this way is safer than changing the server code.
`MobileBackdoorParams` inherits `AppiumParams`. `AppiumParams` has `elementId` member. We should override `elementId` in `MobileBackdoorParams` (but the id in `AppiumParams` final.) I need to confirm and change the `AppiumParams`, too, then.

KazuCocoa:
thanks! @imurchie

PR #457, opened 2019-06-03
 Switch server implementation to use Kotlin

mykola-mokhnach:


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/Server.kt
```diff
@@ -0,0 +1,120 @@
...
+        val files = LinkedHashMap<String, String>()
+        try {
+            session.parseBody(files)
+            response = router!!.route(session.uri, session.method, files)
+        } catch (e: RuntimeException) {
+            response = AppiumResponse(AppiumStatus.UNKNOWN_ERROR, Log.getStackTraceString(e))
+        } catch (e: IOException) {
+            response = AppiumResponse(AppiumStatus.UNKNOWN_ERROR, Log.getStackTraceString(e))
+        } catch (e: ResponseException) {
+            response = AppiumResponse(AppiumStatus.UNKNOWN_ERROR, Log.getStackTraceString(e))
```
KazuCocoa: You can also write like:
```kotlin
        val files = LinkedHashMap<String, String>()
        val response = try {
            session.parseBody(files)
            router!!.route(session.uri, session.method, files)
        } catch (e: Exception) {
            when(e) {
                is RuntimeException, is IOException, is ResponseException ->
                    AppiumResponse(AppiumStatus.UNKNOWN_ERROR, Log.getStackTraceString(e))
                else -> throw e
            }
        }
```

mykola-mokhnach: thanks, changed

PR #473, opened 2019-08-05
Avoid qualifying the activity name if it is a pattern

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/13030

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -8,6 +8,9 @@ import _ from 'lodash';
  * @returns {string} The qualified activity name
  */
 function qualifyActivityName (activityName, packageName) {
+  // if either activity or package name is not set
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -7,7 +8,7 @@
  * @returns {string} The qualified activity name
  */
 function qualifyActivityName (activityName, packageName) {
-  if (!activityName || !packageName) {
+  if ([activityName, packageName].some((name) => !name || _.includes(name, '*'))) {
```
imurchie: While it's clear what this is doing in the context of this PR, I fear it will not make any sense once just a bit of code. A comment would go a long way.

mykola-mokhnach: added a comment

PR #475, opened 2019-08-10
Switch the server and the driver to W3C communication protocol

mykola-mokhnach:
Espresso was always our bleeding edge tech, so lets try to start there and then port the same stuff to other servers in this goes smoothly


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/response/AppiumResponse.kt
```diff
@@ -16,39 +16,47 @@
...
-        this.value = value
-        this.status = status
-        this.sessionId = sessionId
-        id = UUID.randomUUID().toString()
+    private fun formatError(e: AppiumException): Map<String, String> {
+        val result = mutableMapOf<String, String>()
+        result["error"] = e.error()
+        result["message"] = e.message ?: MESSAGE_UNKNOWN_ERROR
+        result["stacktrace"] = Log.getStackTraceString(e)
+        return result
```
KazuCocoa: Can write like:
```kotlin
private fun formatError(e: AppiumException): Map<String, String> = mapOf(
    "error" to e.error(),
    "message" to (e.message ?: MESSAGE_UNKNOWN_ERROR),
    "stacktrace" to Log.getStackTraceString(e)
)
```

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/caps/CapsUtils.kt
```diff
@@ -0,0 +1,100 @@
...
+        result[it.key] = it.value
+    }
+    return result
+}
+
+
+fun stripPrefixes(caps: Map<String, Any?>): Map<String, Any?> {
+    val prefix = "$APPIUM_PREFIX:"
+    val prefixedCaps = caps.keys.filter { it.startsWith(prefix) }
+    val result = caps.toMutableMap()
```
KazuCocoa: Can we remove `result` in this fun?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/http/response/AppiumResponse.kt
mykola-mokhnach: thanks, will change

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/caps/CapsUtils.kt
mykola-mokhnach: removed

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/w3c/caps/CapsUtils.kt
```diff
@@ -0,0 +1,96 @@
...
+const val APPIUM_PREFIX = "appium"
+
+
+fun isStandardCap(capName: String): Boolean = STANDARD_CAPS.any { it.toLowerCase() == capName.toLowerCase() }
+
+
+fun mergeCaps(primary: Map<String, Any?>, secondary: Map<String, Any?>): Map<String, Any?> {
+    val result = primary.toMutableMap()
+    secondary.mapValues {
+        if (result[it.key] != null) {
```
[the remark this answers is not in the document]
mykola-mokhnach: this will throw if `result[it.key]` is null, but we need to do it if it is NOT null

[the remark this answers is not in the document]
KazuCocoa: Ah. Yeah, definitely

dpgraham:
Awesome!

dpgraham:
@mykola-mokhnach thanks for doing this. I've thought quite a bit on how we can consolidate the server logic (just haven't had the time to implement it)

I think we ought to port everything from `uiautomator2-server` into `espresso-server` (and then rename it to something like `androidx-server`) and have 2 separate native contexts.... a UiAutomator2 context and an Espresso context. We could use the automation name to determine which context to use, and even support users switching between uiautomator2 and espresso mid-session (really useful if you're using Espresso driver and need to perform operations outside of the AUT)

PR #485, opened 2019-08-14
fix raise an error for launch app and close app since they do not need in espresso

KazuCocoa:
Current `launchApp` missed espresso session like below scenario since espresso-driver does not establish a new session with the launched app

```
@driver.close_app
assert(@@core.wait { @driver.app_state('io.appium.android.apis') != :running_in_foreground })

@driver.launch_app
e = @@core.wait { @driver.find_element :accessibility_id, 'App' }
assert_equal 'App', e.text
```

Then, below error raises in `find_element`.
```
Appium::Core::Wait::TimeoutError: timed out after 20 seconds (An unknown server-side error occurred while processing the command. Original error: Could not proxy. Proxy error: Could not proxy command to remote server. Original error: Error: socket hang up)
```

https://dev.azure.com/kazucocoa/ruby_lib_core/_build/results?buildId=1265

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -23,12 +23,12 @@ function assertRequiredOptions (options, requiredOptionNames) {
 
 // eslint-disable-next-line require-await
 commands.launchApp = async function launchApp () {
-  throw new errors.UnsupportedOperationError('Please re-create a new session to configure the app under test correctly');
+  throw new errors.UnsupportedOperationError('Please create a new session in order to launch the application under test');
 };
 
 // eslint-disable-next-line require-await
 commands.closeApp = async function closeApp () {
-  throw new errors.UnsupportedOperationError('Please quit the session instead of closeApp');
+  throw new errors.UnsupportedOperationError('Please quit the session in order to close the application under test');
 };
 
 commands.mobilePerformEditorAction = async function mobilePerformEditorAction (opts = {}) {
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
I don't think launchApp/closeApp have much sense in Espresso, since terminating the app means session termination

KazuCocoa:
Ah... yeah, sounds reasonable.
Let me raise an error since they do not work expectedly.
Will mark as not support in espresso driver in https://github.com/appium/appium/blob/master/commands-yml/commands/device/app/launch-app.yml and https://github.com/appium/appium/blob/master/commands-yml/commands/device/app/close-app.yml

lib/commands/general.js
```diff
@@ -21,6 +21,16 @@ function assertRequiredOptions (options, requiredOptionNames) {
     `You have only provided: ${JSON.stringify(presentOptionNames)}`);
 }
 
+// eslint-disable-next-line require-await
+commands.launchApp = async function launchApp () {
+  throw new errors.UnsupportedOperationError('Please re-create a new session to configure the app under test correctly');
```
mykola-mokhnach: Please create a new session in order to launch the application under test

lib/commands/general.js
```diff
@@ -21,6 +21,16 @@ function assertRequiredOptions (options, requiredOptionNames) {
...
+};
+
+// eslint-disable-next-line require-await
+commands.closeApp = async function closeApp () {
+  throw new errors.UnsupportedOperationError('Please quit the session instead of closeApp');
```
mykola-mokhnach: Please quit the session in order to close the application under test

PR #496, opened 2019-09-13
Allow configuring Gradle, AGP, various SDK and Kotlin versions for Espresso server

tinder-ktarasov:
Currently, appium-espresso-driver uses same Gradle, AGP, SDK, and Kotlin versions when building espresso-server for all apps. Such versions might be different from the versions used to build AUT. Such setup is unusual as androidTest APKs are usually built with the same SDKs as the main app. It can lead to various issues as Android building infrastructure is quite complex.
This PR proposes changes that will allow configuring those versions to keep them in sync with the versions in the main app.

To build espresso-server with custom versions **forceEspressoRebuild** capability should be set to true.
I didn't change anything regarding caching, so I suppose rebuilt APK is not cached and it will be rebuilt even when versions are same.

~~Unfortunately, it doesn't fix that Resource issue, I'll make a separate PR for it.~~

**New desired capabilities**
- **showGradleLog** (boolean) - show Gradle build log in the server log
- **espressoBuildConfig** (path as a string) - path to build configuration JSON

**Build configuration JSON keys**
Configuration example:
```json
{
  "toolsVersions": {
    "gradle": "5.1.1",
    "androidGradlePlugin": "3.4.2",
    "compileSdk": 28,
    "buildTools": "28.0.3",
    "minSdk": 18,
    "targetSdk": 28,
    "kotlin": "1.3.31"
  },
  "additionalAppDependencies": [
    "com.google.android.material:material:1.0.0"
  ]
}
```

Minimum supported versions here and default versions are current versions in espresso-server project (not changed in this PR)
- **toolsVersion** - versions of various tools and SDKs used to build espresso-server
    - **gradle** - version of Gradle to use for building espresso-server
    - **androidGradlePlugin** - version of Android Gradle Plugin
    - **buildTools** - Android SDK build tools version
    - **compile_sdk_version** - compileSdk in build.gradle
    - **minSdk** - minSdk in build.gradle
    - **targetSdk** - targetSdk in build.gradle
    - **kotlin** - Kotlin version
- **additionalAppDependencies** - array containing additional dependencies of the tested app that build tools should know about. Items of the array maps to 'implementation' lines in build.gradle. I don't know exact causes but adding 'com.google.android.material:material:1.0.0' here fixes resource related startup issues for some apps.

**Possible improvements (not implemented)**
* Cache espresso-server APKs and reuse them when versions are the same
* Building espresso-server will spawn Gradle daemon process that is not stopped after the build (and will be reused for next builds). Probably some would want to avoid it.

### DIFF
--- a/espresso-server/build.gradle
+++ b/espresso-server/build.gradle
@@ -1,7 +1,25 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.3.31'
+    ext.getIntegerProperty = {String propertyName, int defaultValue ->
+        Object propertyValue = project.getProperties().get(propertyName)
+        if (propertyValue == null) {
+            return defaultValue
+        } else {
+            Integer.parseInt(propertyValue.toString(), 10)
+        }
+    }
+    ext.getStringProperty = { String propertyName, String defaultValue ->
+        Object propertyValue = project.getProperties().get(propertyName)
+        if (propertyValue == null) {
+            return defaultValue
+        } else {
+            propertyValue.toString()
+        }
+    }
+
+    ext.kotlin_version = getStringProperty('kotlin_version', '1.3.31')
+    ext.agp_version = getStringProperty('agp_version', '3.4.2')
 
     repositories {
         maven {
@@ -12,7 +30,7 @@ buildscript {
     }
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath 'com.android.tools.build:gradle:3.4.2'
+        classpath "com.android.tools.build:gradle:$agp_version"
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/espresso-server/build.gradle
+++ b/espresso-server/build.gradle
@@ -3,23 +3,15 @@
 buildscript {
     ext.getIntegerProperty = {String propertyName, int defaultValue ->
         Object propertyValue = project.getProperties().get(propertyName)
-        if (propertyValue == null) {
-            return defaultValue
-        } else {
-            Integer.parseInt(propertyValue.toString(), 10)
-        }
+        (propertyValue) ? Integer.parseInt(propertyValue.toString(), 10) : defaultValue
     }
     ext.getStringProperty = { String propertyName, String defaultValue ->
         Object propertyValue = project.getProperties().get(propertyName)
-        if (propertyValue == null) {
-            return defaultValue
-        } else {
-            propertyValue.toString()
-        }
+        (propertyValue) ? propertyValue.toString() : defaultValue
     }
 
     ext.kotlin_version = getStringProperty('kotlin_version', '1.3.31')
-    ext.agp_version = getStringProperty('agp_version', '3.4.2')
+    ext.android_gradle_plugin_version = getStringProperty('android_gradle_plugin_version', '3.4.2')
 
     repositories {
         maven {
@@ -30,7 +22,7 @@ buildscript {
     }
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "com.android.tools.build:gradle:$agp_version"
+        classpath "com.android.tools.build:gradle:$android_gradle_plugin_version"
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/espresso-server/build.gradle
+++ b/espresso-server/build.gradle
@@ -10,8 +10,8 @@ buildscript {
         (propertyValue) ? propertyValue.toString() : defaultValue
     }
 
-    ext.kotlin_version = getStringProperty('kotlin_version', '1.3.31')
-    ext.android_gradle_plugin_version = getStringProperty('android_gradle_plugin_version', '3.4.2')
+    ext.kotlin_version = getStringProperty('appiumKotlin', '1.3.31')
+    ext.android_gradle_plugin_version = getStringProperty('appiumAndroidGradlePlugin', '3.4.2')
 
     repositories {
         maven {
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import { BaseDriver } from 'appium-base-driver';
+import { BaseDriver, errors, isErrorType } from 'appium-base-driver';
 import { EspressoRunner, TEST_APK_PKG } from './espresso-runner';
 import { fs } from 'appium-support';
 import logger from './logger';
@@ -171,7 +171,13 @@ class EspressoDriver extends BaseDriver {
       return [sessionId, caps];
     } catch (e) {
       await this.deleteSession();
-      throw e;
+      e.message += '. Check https://github.com/appium/appium-espresso-driver#troubleshooting regarding advanced session startup troubleshooting.';
+      if (isErrorType(e, errors.SessionNotCreatedError)) {
+        throw e;
+      }
+      const err = new errors.SessionNotCreatedError(e.message);
+      err.stack = e.stack;
+      throw err;
     }
   }
 
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -345,6 +345,8 @@ class EspressoDriver extends BaseDriver {
       appPackage: this.opts.appPackage,
       appActivity: this.opts.appActivity,
       forceEspressoRebuild: !!this.opts.forceEspressoRebuild,
+      espressoBuildConfig: this.opts.espressoBuildConfig || '',
+      showGradleLog: !!this.opts.showGradleLog,
       serverLaunchTimeout: this.opts.espressoServerLaunchTimeout,
       androidInstallTimeout: this.opts.androidInstallTimeout,
     });
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -345,7 +345,7 @@ class EspressoDriver extends BaseDriver {
       appPackage: this.opts.appPackage,
       appActivity: this.opts.appActivity,
       forceEspressoRebuild: !!this.opts.forceEspressoRebuild,
-      espressoBuildConfig: this.opts.espressoBuildConfig || '',
+      espressoBuildConfig: this.opts.espressoBuildConfig,
       showGradleLog: !!this.opts.showGradleLog,
       serverLaunchTimeout: this.opts.espressoServerLaunchTimeout,
       androidInstallTimeout: this.opts.androidInstallTimeout,
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -23,7 +23,6 @@ const TARGET_PACKAGE_CONTAINER = '/data/local/tmp/espresso.apppackage';
  * @param targetBaseDir {string} directory to copy files to
  */
 async function copyRecursively (sourceBaseDir, targetBaseDir) {
-  logger.debug(`copyRecursively(${sourceBaseDir}, ${targetBaseDir})`);
   await fs.walkDir(sourceBaseDir, true, async (itemPath, isDirectory) => {
     const relativePath = path.relative(sourceBaseDir, itemPath);
     const targetPath = path.resolve(targetBaseDir, relativePath);
@@ -36,7 +35,7 @@ async function copyRecursively (sourceBaseDir, targetBaseDir) {
     if (!isDirectory) {
       await fs.copyFile(itemPath, targetPath);
     } else {
-      mkdirp(targetPath);
+      await mkdirp(targetPath);
     }
     return false;
   });
@@ -138,13 +137,19 @@ class EspressoRunner {
     let versions = {};
     if (this.espressoBuildConfig) {
       logger.info(`Using build configuration JSON from: '${this.espressoBuildConfig}'`);
-      versions = JSON.parse(await fs.readFile(this.espressoBuildConfig));
+      try {
+        versions = JSON.parse(await fs.readFile(this.espressoBuildConfig, 'utf8'));
+      } catch (e) {
+        logger.error('Failed to parse build configuration JSON', e);
+        throw e;
+      }
     }
     logger.info('Building espresso server');
     const serverPath = path.resolve(this.tmpDir, 'espresso-server');
     logger.debug(`build dir: ${serverPath}`);
     await fs.rimraf(serverPath);
     await mkdirp(serverPath);
+    logger.debug(`Copying espresso server template from (${TEST_SERVER_PATH} to ${serverPath})`);
     await copyRecursively(TEST_SERVER_PATH, serverPath);
     await new ServerBuilder({serverPath, versions, showGradleLog: this.showGradleLog}).build();
     const apkPath = path.resolve(serverPath, 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -12,7 +12,7 @@ import { copyGradleProjectRecursively } from './utils';
 
 
 const TEST_SERVER_ROOT = path.resolve(__dirname, '..', '..', 'espresso-server');
-const TEST_MANIFEST_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'AndroidManifest-test.xml');
+const TEST_MANIFEST_PATH = path.resolve(TEST_SERVER_ROOT, 'AndroidManifest-test.xml');
 const TEST_APK_PKG = 'io.appium.espressoserver.test';
 const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild'];
 const ESPRESSO_SERVER_LAUNCH_TIMEOUT = 30000;
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -14,7 +14,7 @@ import { copyRecursively } from './utils';
 const TEST_SERVER_PATH = path.resolve(__dirname, '..', '..', 'espresso-server');
 const TEST_MANIFEST_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'AndroidManifest-test.xml');
 const TEST_APK_PKG = 'io.appium.espressoserver.test';
-const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild', 'espressoBuildConfig'];
+const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild'];
 const ESPRESSO_SERVER_LAUNCH_TIMEOUT = 30000;
 const TARGET_PACKAGE_CONTAINER = '/data/local/tmp/espresso.apppackage';
 
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -11,7 +11,7 @@ import _ from 'lodash';
 import { copyRecursively } from './utils';
 
 
-const TEST_SERVER_PATH = path.resolve(__dirname, '..', '..', 'espresso-server');
+const TEST_SERVER_ROOT = path.resolve(__dirname, '..', '..', 'espresso-server');
 const TEST_MANIFEST_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'AndroidManifest-test.xml');
 const TEST_APK_PKG = 'io.appium.espressoserver.test';
 const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild'];
@@ -126,8 +126,8 @@ class EspressoRunner {
     logger.debug(`build dir: ${serverPath}`);
     await fs.rimraf(serverPath);
     await mkdirp(serverPath);
-    logger.debug(`Copying espresso server template from (${TEST_SERVER_PATH} to ${serverPath})`);
-    await copyRecursively(TEST_SERVER_PATH, serverPath);
+    logger.debug(`Copying espresso server template from (${TEST_SERVER_ROOT} to ${serverPath})`);
+    await copyRecursively(TEST_SERVER_ROOT, serverPath);
     await new ServerBuilder({serverPath, versions, showGradleLog: this.showGradleLog}).build();
     const apkPath = path.resolve(serverPath, 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
 
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -8,6 +8,7 @@ import { version } from '../../package.json'; // eslint-disable-line import/no-u
 import request from 'request-promise';
 import B from 'bluebird';
 import _ from 'lodash';
+import { copyRecursively } from './utils';
 
 
 const TEST_SERVER_PATH = path.resolve(__dirname, '..', '..', 'espresso-server');
@@ -16,30 +17,6 @@ const TEST_APK_PKG = 'io.appium.espressoserver.test';
 const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild', 'espressoBuildConfig'];
 const ESPRESSO_SERVER_LAUNCH_TIMEOUT = 30000;
 const TARGET_PACKAGE_CONTAINER = '/data/local/tmp/espresso.apppackage';
-
-/**
- * Recursively copy all files
- * @param sourceBaseDir {string} directory to copy files from
- * @param targetBaseDir {string} directory to copy files to
- */
-async function copyRecursively (sourceBaseDir, targetBaseDir) {
-  await fs.walkDir(sourceBaseDir, true, async (itemPath, isDirectory) => {
-    const relativePath = path.relative(sourceBaseDir, itemPath);
-    const targetPath = path.resolve(targetBaseDir, relativePath);
-
-    const isGradleBuildDir = `${path.sep}${itemPath}${path.sep}`.includes(`${path.sep}build${path.sep}`);
-    if (isGradleBuildDir) {
-      return false;
-    }
-
-    if (!isDirectory) {
-      await fs.copyFile(itemPath, targetPath);
-    } else {
-      await mkdirp(targetPath);
-    }
-    return false;
-  });
-}
 
 class EspressoRunner {
   constructor (opts = {}) {
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -8,7 +8,7 @@ import { version } from '../../package.json'; // eslint-disable-line import/no-u
 import request from 'request-promise';
 import B from 'bluebird';
 import _ from 'lodash';
-import { copyRecursively } from './utils';
+import { copyGradleProjectRecursively } from './utils';
 
 
 const TEST_SERVER_ROOT = path.resolve(__dirname, '..', '..', 'espresso-server');
@@ -127,7 +127,7 @@ class EspressoRunner {
     await fs.rimraf(serverPath);
     await mkdirp(serverPath);
     logger.debug(`Copying espresso server template from (${TEST_SERVER_ROOT} to ${serverPath})`);
-    await copyRecursively(TEST_SERVER_ROOT, serverPath);
+    await copyGradleProjectRecursively(TEST_SERVER_ROOT, serverPath);
     await new ServerBuilder({serverPath, versions, showGradleLog: this.showGradleLog}).build();
     const apkPath = path.resolve(serverPath, 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
 
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -1,6 +1,7 @@
 import { JWProxy } from 'appium-base-driver';
 import { retryInterval } from 'asyncbox';
 import logger from './logger';
+import ServerBuilder from './server-builder';
 import path from 'path';
 import { fs, util, mkdirp } from 'appium-support';
 import { version } from '../../package.json'; // eslint-disable-line import/no-unresolved
@@ -9,12 +10,37 @@ import B from 'bluebird';
 import _ from 'lodash';
 
 
-const TEST_APK_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
+const TEST_SERVER_PATH = path.resolve(__dirname, '..', '..', 'espresso-server');
 const TEST_MANIFEST_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'AndroidManifest-test.xml');
 const TEST_APK_PKG = 'io.appium.espressoserver.test';
-const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild'];
+const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild', 'espressoBuildConfig'];
 const ESPRESSO_SERVER_LAUNCH_TIMEOUT = 30000;
 const TARGET_PACKAGE_CONTAINER = '/data/local/tmp/espresso.apppackage';
+
+/**
+ * Recursively copy all files
+ * @param sourceBaseDir {string} directory to copy files from
+ * @param targetBaseDir {string} directory to copy files to
+ */
+async function copyRecursively (sourceBaseDir, targetBaseDir) {
+  logger.debug(`copyRecursively(${sourceBaseDir}, ${targetBaseDir})`);
+  await fs.walkDir(sourceBaseDir, true, async (itemPath, isDirectory) => {
+    const relativePath = path.relative(sourceBaseDir, itemPath);
+    const targetPath = path.resolve(targetBaseDir, relativePath);
+
+    const isGradleBuildDir = `${path.sep}${itemPath}${path.sep}`.includes(`${path.sep}build${path.sep}`);
+    if (isGradleBuildDir) {
+      return false;
+    }
+
+    if (!isDirectory) {
+      await fs.copyFile(itemPath, targetPath);
+    } else {
+      mkdirp(targetPath);
+    }
+    return false;
+  });
+}
 
 class EspressoRunner {
   constructor (opts = {}) {
@@ -28,6 +54,8 @@ class EspressoRunner {
     this.proxyReqRes = this.jwproxy.proxyReqRes.bind(this.jwproxy);
 
     this.modServerPath = path.resolve(this.tmpDir, `${TEST_APK_PKG}_${version}_${this.appPackage}.apk`);
+    this.showGradleLog = opts.showGradleLog;
+    this.espressoBuildConfig = opts.espressoBuildConfig;
 
     this.serverLaunchTimeout = opts.serverLaunchTimeout || ESPRESSO_SERVER_LAUNCH_TIMEOUT;
     this.androidInstallTimeout = opts.androidInstallTimeout;
@@ -107,6 +135,20 @@ class EspressoRunner {
   }
 
   async buildNewModServer () {
+    let versions = {};
+    if (this.espressoBuildConfig) {
+      logger.info(`Using build configuration JSON from: '${this.espressoBuildConfig}'`);
+      versions = JSON.parse(await fs.readFile(this.espressoBuildConfig));
+    }
+    logger.info('Building espresso server');
+    const serverPath = path.resolve(this.tmpDir, 'espresso-server');
+    logger.debug(`build dir: ${serverPath}`);
+    await fs.rimraf(serverPath);
+    await mkdirp(serverPath);
+    await copyRecursively(TEST_SERVER_PATH, serverPath);
+    await new ServerBuilder({serverPath, versions, showGradleLog: this.showGradleLog}).build();
+    const apkPath = path.resolve(serverPath, 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
+
     logger.info(`Repackaging espresso server for: '${this.appPackage}'`);
     const packageTmpDir = path.resolve(this.tmpDir, this.appPackage);
     const newManifestPath = path.resolve(this.tmpDir, 'AndroidManifest.xml');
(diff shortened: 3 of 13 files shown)


espresso-server/build.gradle
```diff
@@ -1,7 +1,25 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.3.31'
+    ext.getIntegerProperty = {String propertyName, int defaultValue ->
+        Object propertyValue = project.getProperties().get(propertyName)
+        if (propertyValue == null) {
+            return defaultValue
+        } else {
```
mykola-mokhnach: how about using the [elvis](https://mrhaki.blogspot.com/2009/08/groovy-goodness-elvis-operator.html) operator instead?

espresso-server/build.gradle
```diff
@@ -1,7 +1,25 @@
...
+            propertyValue.toString()
+        }
+    }
+
+    ext.kotlin_version = getStringProperty('kotlin_version', '1.3.31')
+    ext.agp_version = getStringProperty('agp_version', '3.4.2')
```
mykola-mokhnach: what `agp` means?

lib/espresso-runner.js
```diff
@@ -9,13 +10,38 @@ import B from 'bluebird';
...
 const ESPRESSO_SERVER_LAUNCH_TIMEOUT = 30000;
 const TARGET_PACKAGE_CONTAINER = '/data/local/tmp/espresso.apppackage';
 
+/**
+ * Recursively copy all files
+ * @param sourceBaseDir {string} directory to copy files from
+ * @param targetBaseDir {string} directory to copy files to
+ */
+async function copyRecursively (sourceBaseDir, targetBaseDir) {
+  logger.debug(`copyRecursively(${sourceBaseDir}, ${targetBaseDir})`);
```
mykola-mokhnach: I'd prefer human-readable log string ;)

lib/espresso-runner.js
```diff
@@ -9,13 +10,38 @@ import B from 'bluebird';
...
+
+    const isGradleBuildDir = `${path.sep}${itemPath}${path.sep}`.includes(`${path.sep}build${path.sep}`);
+    if (isGradleBuildDir) {
+      return false;
+    }
+
+    if (!isDirectory) {
+      await fs.copyFile(itemPath, targetPath);
+    } else {
+      mkdirp(targetPath);
```
mykola-mokhnach: await

lib/driver.js
```diff
@@ -345,6 +345,8 @@ class EspressoDriver extends BaseDriver {
       appPackage: this.opts.appPackage,
       appActivity: this.opts.appActivity,
       forceEspressoRebuild: !!this.opts.forceEspressoRebuild,
+      espressoBuildConfig: this.opts.espressoBuildConfig || '',
```
mykola-mokhnach: is it necessary to default this variable to an empty string?

lib/espresso-runner.js
```diff
@@ -107,6 +135,20 @@ class EspressoRunner {
   }
 
   async buildNewModServer () {
+    let versions = {};
+    if (this.espressoBuildConfig) {
+      logger.info(`Using build configuration JSON from: '${this.espressoBuildConfig}'`);
+      versions = JSON.parse(await fs.readFile(this.espressoBuildConfig));
```
mykola-mokhnach: there should be a proper human-readable message here if the file content is not a valid JSON

mykola-mokhnach: readFile returns buffer by default. Provide the actual encoding as the second parameter to turn the result to a string

lib/espresso-runner.js
```diff
@@ -9,13 +10,38 @@ import B from 'bluebird';
...
+const REQUIRED_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'appPackage', 'forceEspressoRebuild', 'espressoBuildConfig'];
 const ESPRESSO_SERVER_LAUNCH_TIMEOUT = 30000;
 const TARGET_PACKAGE_CONTAINER = '/data/local/tmp/espresso.apppackage';
 
+/**
+ * Recursively copy all files
+ * @param sourceBaseDir {string} directory to copy files from
+ * @param targetBaseDir {string} directory to copy files to
+ */
+async function copyRecursively (sourceBaseDir, targetBaseDir) {
```
mykola-mokhnach: consider moving this function to utils.js

mykola-mokhnach: it would be nice to have an integration test for it

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+  'build_tools_version',
+  'min_sdk_version',
+  'target_sdk_version',
+  'kotlin_version'
+];
+
+const gradleLog = logger.getLogger('Gradle');
+
+class ServerBuilder {
+  server_versions
```
mykola-mokhnach: what is the purpose of this variable?

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+
+  constructor (args = {}) {
+    this.serverPath = args.serverPath;
+    this.showGradleLog = args.showGradleLog;
+
+    const versions_arg = args.versions || {};
```
mykola-mokhnach: Please use javascript naming style (camelCase)

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+      await this.setGradleWrapperVersion(this.server_versions[GRADLE_VERSION_KEY]);
+    }
+    await this.runBuildProcess();
+  }
+
+  getCommand () {
+    let cmd = `.${path.sep}gradlew`;
+    let args = [];
+
+    for (const key of VERSION_KEYS) {
```
mykola-mokhnach: we could use map/filter here

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+    const {cmd, args} = this.getCommand();
+    log.debug(`Beginning build with command '${cmd} ${args.join(' ')}' ` +
+      `in directory '${this.serverPath}'`);
+    const gradlebuild = new SubProcess(cmd, args, {
+      cwd: this.serverPath,
+      env: process.env,
+      detached: true,
+      stdio: ['ignore', 'pipe', 'pipe'],
+    });
+    gradlebuild._last_lines = [];
```
mykola-mokhnach: `lastLines`

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+
+    const logGradleOutput = this.showGradleLog;
```
mykola-mokhnach: why do we need this variable?

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+    const logMsg = _.isBoolean(this.showGradleLog)
```
mykola-mokhnach: will it work as expected? I saw that !!is applied above to this option

lib/server-builder.js
mykola-mokhnach: why should this variable be attached to gradlebuild ?

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+
+    gradlebuild._last_lines = [];
+
+    // wrap the start procedure in a promise so that we can catch, and report,
+    // any startup errors that are thrown as events
+    return new B((resolve, reject) => {
+      gradlebuild.on('exit', (code, signal) => {
+        log.error(`Gradle exited with code '${code}' and signal '${signal}'`);
+
+        gradlebuild.processExited = true;
```
mykola-mokhnach: why this is needed?

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+      stdio: ['ignore', 'pipe', 'pipe'],
+    });
+    gradlebuild._last_lines = [];
+
+    const logGradleOutput = this.showGradleLog;
+    const logMsg = _.isBoolean(this.showGradleLog)
+      ? `Output from Gradle ${this.showGradleLog ? 'will' : 'will not'} be logged`
+      : 'Output from Gradle will only be logged if any errors are present there';
+    log.debug(`${logMsg}. To change this, use 'showGradleLog' desired capability`);
+    gradlebuild.on('output', (stdout, stderr) => {
```
mykola-mokhnach: consider listening to `lines-${streamName}` or `stream-line` messages. See https://github.com/appium/node-teen_process/blob/e06cf827cd428541ff972654406d788481909575/lib/subprocess.js#L45 for more details

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  runBuildProcess () {
+    const {cmd, args} = this.getCommand();
+    log.debug(`Beginning build with command '${cmd} ${args.join(' ')}' ` +
+      `in directory '${this.serverPath}'`);
+    const gradlebuild = new SubProcess(cmd, args, {
+      cwd: this.serverPath,
+      env: process.env,
+      detached: true,
```
mykola-mokhnach: any reason to make the process detached?

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+      .join('\n');
+  }
+
+  runBuildProcess () {
+    const {cmd, args} = this.getCommand();
+    log.debug(`Beginning build with command '${cmd} ${args.join(' ')}' ` +
+      `in directory '${this.serverPath}'`);
+    const gradlebuild = new SubProcess(cmd, args, {
+      cwd: this.serverPath,
+      env: process.env,
```
mykola-mokhnach: Isn't this the default setting?

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+        if (line.startsWith(GRADLE_URL_PREFIX)) {
+          return GRADLE_URL_PREFIX + GRADLE_URL_TEMPLATE.replace('VERSION', version);
+        } else {
+          return line;
+        }
+      })
+      .join('\n');
+  }
+
+  runBuildProcess () {
```
mykola-mokhnach: async

mykola-mokhnach:
We will need the feature and all the properties documented properly. https://github.com/appium/appium/tree/master/docs/en/writing-running-appium/android would be a good place to keep the document.

Altogether, I like the implementation, good job.

espresso-server/build.gradle
tinder-ktarasov: Android Gradle plugin

lib/espresso-runner.js
tinder-ktarasov: Actually I forgot to remove this log line. Do you think it makes sense to leave it and make more readable?

lib/server-builder.js
tinder-ktarasov: this is just explicit property declaration for syntax highlighting purposes. I suppose Appium style is to not declare them, so I'll remove it and review my code for similar small issues.

lib/server-builder.js
```diff
@@ -0,0 +1,154 @@
...
+
+  async build () {
+    if (this.server_versions[GRADLE_VERSION_KEY]) {
+      await this.setGradleWrapperVersion(this.server_versions[GRADLE_VERSION_KEY]);
+    }
+    await this.runBuildProcess();
+  }
+
+  getCommand () {
+    let cmd = `.${path.sep}gradlew`;
```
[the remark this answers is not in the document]
tinder-ktarasov: I didn't check the driver as whole on Windows, but running `.\gradlew` in command-line interpreter for Android project works on Windows.

lib/server-builder.js
tinder-ktarasov: Here I've just copied approach from xcuitest driver. I don't think there are any issues with having two separate variables.

espresso-server/build.gradle
mykola-mokhnach: I'd prefer to use full name instead of abbreviation then. It's not very common

lib/espresso-runner.js
mykola-mokhnach: yes, everything that helps to track the flow is helpful

lib/server-builder.js
mykola-mokhnach: we simply try to avoid static variables. Moving it to module level could be a good idea. We could also use LRU cache to make sure the value is never going to be too large

lib/server-builder.js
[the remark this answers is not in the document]
mykola-mokhnach: you should make sure it works from `cmd` command line. I suppose in windows the script has even different name (with `.bat` extension)

KazuCocoa:
Awesome!

And it's good to add a description about the caps in https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/caps.md after finalising this change.

tinder-ktarasov:
I've done some additional testing and found that our app require additional dependency on `com.google.android.material:material:1.0.0` in espresso-server. So I'd like to propose this configuration format:
```
{
  "toolsVersions": {
    "gradle": "5.1.1",
    "androidGradlePlugin": "3.4.2",
    "compileSdk": 28,
    "buildTools": "28.0.3",
    "minSdk": 18,
    "targetSdk": 28,
    "kotlin": "1.3.31"
  }
  "additionalAppDependencies": [// maps to 'implementation' lines in build.gradle
    "com.google.android.material:material:1.0.0"
  ]
}
```
What do you think?

KazuCocoa:
ah, nice. thanks :)
Will it allow to `exclude` or `include` as `additionalAppDependencies`? Or simply just add dependencies?

(I think only adding them is enough so far.)

lib/server-builder.js
mykola-mokhnach: it will work if you add await to `new B`
async modifier makes it easier to understand the function returns a promise

espresso-server/build.gradle
tinder-ktarasov: I don't think using elvis operator is good idea here, so I replaced this block with ternary operator.

lib/server-builder.js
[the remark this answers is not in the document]
tinder-ktarasov: Good point. Even though `.\gradlew` worked for me from cmd shell, it doesn't work from SubProcess.

lib/server-builder.js
tinder-ktarasov: I've reworked this function to avoid `detach` and it is much more readable now, also no `new B` now.

lib/espresso-runner.js
tinder-ktarasov: How would you implement such a test? By stubbing fs functions?

tinder-ktarasov:
@mykola-mokhnach, I've addressed most of your comments. So there are 2 things left to do:
- implementing `additionalAppDependencies` suggestion above
- test for `copyRecursively`. I need some help with it - what is the best way to approach it and probably some example on stubbing fs functions (if such stubs are needed)

mykola-mokhnach:
> @mykola-mokhnach, I've addressed most of your comments. So there are 2 things left to do:

Thanks, nicely done

> test for copyRecursively. I need some help with it - what is the best way to approach it and probably some example on stubbing fs functions (if such stubs are needed)

I'm talking about a simple functional test. Create a temporary folder, put some subfolder with files/folders there, then copy the content to another temp folder and verify the copied content is there (files and (sub)folders).

lib/server-builder.js
```diff
@@ -0,0 +1,124 @@
...
+    this.showGradleLog = args.showGradleLog;
+
+    const versionsArg = args.versions || {};
+    const serverVersions = {};
+    for (const key of VERSION_KEYS) {
+      if (versionsArg[key]) {
+        serverVersions[key] = versionsArg[key];
+      }
+    }
+    this.server_versions = serverVersions;
```
mykola-mokhnach: Please use camelCase naming

lib/server-builder.js
```diff
@@ -0,0 +1,124 @@
...
+
+  async build () {
+    if (this.server_versions[GRADLE_VERSION_KEY]) {
+      await this.setGradleWrapperVersion(this.server_versions[GRADLE_VERSION_KEY]);
+    }
+    await this.runBuildProcess();
+  }
+
+  getCommand () {
+    let cmd = system.isWindows() ? 'gradlew.bat' : './gradlew';
```
mykola-mokhnach: 👍

mykola-mokhnach: const

lib/server-builder.js
```diff
@@ -0,0 +1,124 @@
...
+    let args = VERSION_KEYS
+      .filter(key => key !== GRADLE_VERSION_KEY)
+      .map(key => {
+        const serverVersion = this.server_versions[key];
+        return serverVersion ? `-P${key}=${serverVersion}` : null;
+      })
+      .filter(serverVersion => serverVersion);
```
mykola-mokhnach: filter(Boolean)

lib/server-builder.js
```diff
@@ -0,0 +1,124 @@
...
+    const originalProperties = await fs.readFile(propertiesPath, 'utf8');
+    const newProperties = this.updateGradleDistUrl(originalProperties, version);
+    await fs.writeFile(propertiesPath, newProperties, 'utf8');
+  }
+
+  updateGradleDistUrl (propertiesContent, version) {
+    return propertiesContent.split('\n')
+      .map(function (line) {
+        if (line.startsWith(GRADLE_URL_PREFIX)) {
+          return GRADLE_URL_PREFIX + GRADLE_URL_TEMPLATE.replace('VERSION', version);
```
mykola-mokhnach: use 

```
`${GRADLE_URL_PREFIX}{GRADLE_URL_TEMPLATE.replace('VERSION', version)}`
```

lib/server-builder.js
```diff
@@ -0,0 +1,124 @@
...
+
+    buildLastLines = [];
+
+    try {
+      await gradlebuild.start();
+      await gradlebuild.join();
+    } catch (err) {
+      let msg = `Unable to build Espresso server - ${err}\n` +
+        `Gradle error message:${EOL}${buildLastLines}`;
+      log.error(msg);
```
mykola-mokhnach: there is `log.errorAndThrow` method for this purpose

test/unit/server-builder-specs.js
```diff
@@ -0,0 +1,62 @@
+import chai from 'chai';
```
mykola-mokhnach: 👍

test/unit/server-builder-specs.js
```diff
@@ -0,0 +1,62 @@
...
+
+    it('should not pass properties when no versions are specified', function () {
+      const expected = {cmd: expected_cmd, args: ['assembleAndroidTest']};
+      new ServerBuilder().getCommand().should.deep.eql(expected);
+    });
+
+    it('should pass only specified versions as properties', function () {
+      const expected = {cmd: expected_cmd, args: ['-Pandroid_gradle_plugin_version=1.2.3', 'assembleAndroidTest']};
+      let serverBuilder = new ServerBuilder({versions: {android_gradle_plugin_version: '1.2.3'}});
+      serverBuilder.getCommand().should.deep.eql(expected);
```
mykola-mokhnach: eql already does deep comparison

lib/espresso-runner.js
```diff
@@ -1,15 +1,17 @@
...
 import { fs, util, mkdirp } from 'appium-support';
 import { version } from '../../package.json'; // eslint-disable-line import/no-unresolved
 import request from 'request-promise';
 import B from 'bluebird';
 import _ from 'lodash';
+import { copyRecursively } from './utils';
 
 
-const TEST_APK_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
+const TEST_SERVER_PATH = path.resolve(__dirname, '..', '..', 'espresso-server');
```
mykola-mokhnach: how about TEST_SERVER_ROOT?

lib/server-builder.js
mykola-mokhnach: Also I assume that `./` prefix is not needed for gradlew

tinder-ktarasov: As I understand it, `./` prefix makes sure that gradlew will be started from the passed working directory, not from somewhere else in PATH. I think this would be more clean than going without `./`.

mykola-mokhnach: not really. for a normal command prompt adding `./` prefix requests to execute the file
For subprocess having `cwd` option set is enough though

tinder-ktarasov: Unfortunately just `gradlew` doesn't work on my mac, so I'll keep it `./gradlew`. For some reason it tries to run `gradlew` from PATH if there is no `./`.

mykola-mokhnach: hm, ok. Thanks for checking it

test/functional/utils-specs.js
```diff
@@ -0,0 +1,87 @@
...
+const expect = chai.expect;
+
+describe('copyGradleProjectRecursively', function () {
+  let baseSrcDir;
+  let baseDestDir;
+
+  function expectNotExist (dir) {
+    expect(function () {
+      fs.accessSync(dir, fs.constants.F_OK);
+    }).to.throw();
```
tinder-ktarasov: I wasn't able to write this check using `async/await`+`chai-as-promised`, so I had to you sync version of the function. It is ok for me, if someone can rewrite it to be async.

test/functional/utils-specs.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  beforeEach(function () {
+    baseSrcDir = nodefs.mkdtempSync(path.join(os.tmpdir(), 'appium-src-'));
+    baseDestDir = nodefs.mkdtempSync(path.join(os.tmpdir(), 'appium-dst-'));
+  });
+
+  afterEach(async function () {
+    await fs.rimraf(baseSrcDir);
+    await fs.rimraf(baseDestDir);
```
tinder-ktarasov: My understanding of `rimraf` docs is that this method doesn't fail when directory doesn't exist. Could someone confirm this?

lib/server-builder.js
mykola-mokhnach: bump

test/functional/utils-specs.js
mykola-mokhnach: use `promiseFunc.should.eventually.be.rejected/rejectedWith(/pattern/)`

test/functional/utils-specs.js
mykola-mokhnach: yes it does not

test/functional/utils-specs.js
```diff
@@ -0,0 +1,87 @@
...
+  async function expectCorrectFileContentIn (filepath) {
+    await fs.readFile(filepath, 'utf8').should.eventually.eql('foobar');
+  }
+
+  async function createTestFile (filepath) {
+    await fs.writeFile(filepath, 'foobar', 'utf8');
+  }
+
+  beforeEach(function () {
+    baseSrcDir = nodefs.mkdtempSync(path.join(os.tmpdir(), 'appium-src-'));
```
mykola-mokhnach: use tempDir.openDir from appium-support module

lib/espresso-runner.js
```diff
@@ -1,15 +1,17 @@
...
 import { version } from '../../package.json'; // eslint-disable-line import/no-unresolved
 import request from 'request-promise';
 import B from 'bluebird';
 import _ from 'lodash';
+import { copyGradleProjectRecursively } from './utils';
 
 
-const TEST_APK_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
+const TEST_SERVER_ROOT = path.resolve(__dirname, '..', '..', 'espresso-server');
 const TEST_MANIFEST_PATH = path.resolve(__dirname, '..', '..', 'espresso-server', 'AndroidManifest-test.xml');
```
mykola-mokhnach: this values could be replaces with path.resolve(TEST_SERVER_ROOT, 'AndroidManifest-test.xml')

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+    const buildConfiguration = args.buildConfiguration || {};
+
+    const versionConfiguration = buildConfiguration.toolsVersions || {};
+    const serverVersions = {};
+    for (const key of VERSION_KEYS) {
+      if (versionConfiguration[key]) {
+        serverVersions[key] = versionConfiguration[key];
+      }
+    }
+    this.serverVersions = serverVersions;
```
mykola-mokhnach: this could nicely use VERSION_KEYS.reduce ...

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+
+    this.additionalAppDependencies = buildConfiguration.additionalAppDependencies || [];
+  }
+
+  async build () {
+    if (this.serverVersions[GRADLE_VERSION_KEY]) {
+      await this.setGradleWrapperVersion(this.serverVersions[GRADLE_VERSION_KEY]);
+    }
+    await this.insertAdditionalAppDependencies(this.additionalAppDependencies);
```
mykola-mokhnach: should we check if this.additionalAppDependencies is not empty first?

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+
+  async setGradleWrapperVersion (version) {
+    const propertiesPath = path.resolve(this.serverPath, 'gradle', 'wrapper', 'gradle-wrapper.properties');
+    const originalProperties = await fs.readFile(propertiesPath, 'utf8');
+    const newProperties = this.updateGradleDistUrl(originalProperties, version);
+    await fs.writeFile(propertiesPath, newProperties, 'utf8');
+  }
+
+  updateGradleDistUrl (propertiesContent, version) {
+    return propertiesContent.split('\n')
```
mykola-mokhnach: I would rather replace it by regexp

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+    args.push('assembleAndroidTest');
+
+    return {cmd, args};
+  }
+
+  async setGradleWrapperVersion (version) {
+    const propertiesPath = path.resolve(this.serverPath, 'gradle', 'wrapper', 'gradle-wrapper.properties');
+    const originalProperties = await fs.readFile(propertiesPath, 'utf8');
+    const newProperties = this.updateGradleDistUrl(originalProperties, version);
+    await fs.writeFile(propertiesPath, newProperties, 'utf8');
```
mykola-mokhnach: should we check if newProperties is different from originalProperties first?

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  updateDependencyLines (configurationContent, additionalAppDependencies) {
+    if (!Array.isArray(additionalAppDependencies)) {
+      throw new Error('additionalAppDependencies must be an array');
+    }
+    const dependencyLines = additionalAppDependencies
+      .map(function (dependency) {
+        if (/[\s']/.test(dependency)) {
+          throw new Error(`Disallowed characters in an additional dependency: ${dependency}`);
```
mykola-mokhnach: we probably would like to show which chars are not acceptable in the error message

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+        }
+        return `implementation '${dependency}'`;
+      })
+      .join('\n');
+
+    return configurationContent.split('\n')
```
mykola-mokhnach: again, would prefer replacement by regexp

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+      if (this.showGradleLog && !ignoreError) {
+        gradleLog.error(line);
+      }
+      buildLastLines.push(`${EOL}${line}`);
+      if (buildLastLines.length > GRADLE_MAX_ERROR_LOG_LINES) {
+        buildLastLines = buildLastLines.slice(-GRADLE_MAX_ERROR_LOG_LINES);
+      }
+    });
+
+    buildLastLines = [];
```
mykola-mokhnach: why it should be assigned again?

lib/server-builder.js
```diff
@@ -0,0 +1,161 @@
...
+
+    try {
+      await gradlebuild.start();
+      await gradlebuild.join();
+    } catch (err) {
+      let msg = `Unable to build Espresso server - ${err}\n` +
+        `Gradle error message:${EOL}${buildLastLines}`;
```
mykola-mokhnach: should we really collect these lines if we have `showGradleLog`? I would just suggest setting it to true if it is not set and checking the server log.

lib/utils.js
```diff
@@ -22,4 +24,28 @@ function qualifyActivityName (activityName, packageName) {
...
+  await fs.walkDir(sourceBaseDir, true, async (itemPath, isDirectory) => {
+    const relativePath = path.relative(sourceBaseDir, itemPath);
+    const targetPath = path.resolve(targetBaseDir, relativePath);
+
+    const isInGradleBuildDir = `${path.sep}${itemPath}`.includes(`${path.sep}build${path.sep}`);
+    if (isInGradleBuildDir) {
+      return false;
+    }
+
+    if (!isDirectory) {
```
mykola-mokhnach: nit: positive conditions are easier to read

test/functional/utils-specs.js
```diff
@@ -0,0 +1,82 @@
...
+
+chai.should();
+chai.use(chaiAsPromised);
+
+describe('copyGradleProjectRecursively', function () {
+  let baseSrcDir;
+  let baseDestDir;
+
+  function expectNotExist (file) {
+    fs.access(file, fs.constants.F_OK).should.eventually.be.rejectedWith(/no such file/);
```
mykola-mokhnach: await should still be there

test/functional/utils-specs.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  function expectCorrectFileContentIn (filepath) {
+    fs.readFile(filepath, 'utf8').should.eventually.eql('foobar');
```
mykola-mokhnach: same here

lib/server-builder.js
tinder-ktarasov: It is not required as current implementation should not break anything. But yes, that will remove some IO calls when they are not required.

lib/server-builder.js
tinder-ktarasov: This line will only be executed if some Gradle version was specified in a build configuration file. So I'm not sure such check is needed.

lib/server-builder.js
tinder-ktarasov: Ok. Probably `/[\r\n]/`?

lib/server-builder.js
tinder-ktarasov: That will complicate debugging flaky issues with building. So I would still like to have output for build failures, at least by default.
Also I think XCUITest driver prints error log when similar capability is `false`.

tinder-ktarasov:
Signed the CLA

lib/server-builder.js
```diff
@@ -32,13 +32,17 @@ class ServerBuilder {
...
 
     const versionConfiguration = buildConfiguration.toolsVersions || {};
-    const serverVersions = {};
-    for (const key of VERSION_KEYS) {
-      if (versionConfiguration[key]) {
-        serverVersions[key] = versionConfiguration[key];
-      }
-    }
-    this.serverVersions = serverVersions;
+    this.serverVersions = Object.entries(versionConfiguration)
```
mykola-mokhnach: the _reduce function for lodash lib could make it all much shorter. See https://github.com/appium/appium-idb/blob/2dce71d2b21b5d6bf95fcd5bf74e4c6bef0549ea/lib/helpers.js#L56

lib/server-builder.js
```diff
@@ -47,7 +51,14 @@ class ServerBuilder {
     if (this.serverVersions[GRADLE_VERSION_KEY]) {
       await this.setGradleWrapperVersion(this.serverVersions[GRADLE_VERSION_KEY]);
     }
-    await this.insertAdditionalAppDependencies(this.additionalAppDependencies);
+
+    if (!Array.isArray(this.additionalAppDependencies)) {
+      throw new Error('additionalAppDependencies must be an array');
+    }
+    if (this.additionalAppDependencies.length > 0) {
```
mykola-mokhnach: you could use _.isEmpty lodash function

lib/server-builder.js
```diff
@@ -75,7 +86,7 @@ class ServerBuilder {
   }
 
   updateGradleDistUrl (propertiesContent, version) {
-    return propertiesContent.split('\n')
+    return propertiesContent.split(/\n/)
```
mykola-mokhnach: I was talking about 

```
propertiesContent.replace(new RegExp(`^${_escapeRegex(GRADLE_URL_PREFIX)}`, 'gm'), `$1${GRADLE_URL_TEMPLATE.replace('VERSION', version)}`)
```

lib/server-builder.js
```diff
@@ -94,19 +105,16 @@ class ServerBuilder {
...
 
   updateDependencyLines (configurationContent, additionalAppDependencies) {
-    if (!Array.isArray(additionalAppDependencies)) {
-      throw new Error('additionalAppDependencies must be an array');
-    }
     const dependencyLines = additionalAppDependencies
       .map(function (dependency) {
         if (/[\s']/.test(dependency)) {
-          throw new Error(`Disallowed characters in an additional dependency: ${dependency}`);
+          throw new Error(`Single quotes and space characters are disallowed in additional dependencies: ${dependency}`);
```
mykola-mokhnach: I would say `space` -> `whitespace`, since `\s` also matches tabs, line breaks, etc

lib/server-builder.js
```diff
@@ -94,19 +105,16 @@ class ServerBuilder {
...
         }
         return `implementation '${dependency}'`;
       })
       .join('\n');
 
-    return configurationContent.split('\n')
+    return configurationContent.split(/\n/)
```
mykola-mokhnach: same as above

hunzai:
@mykola-mokhnach @tinder-ktarasov let me know if I can help testing, at least :)

lib/server-builder.js
```diff
@@ -32,17 +33,14 @@ class ServerBuilder {
...
-        if (!VERSION_KEYS.includes(key)) {
-          log.warn(`Got unexpected '${key}' in toolsVersion block of the build configuration`);
-          return false;
-        }
-        return true;
-      })
-      .reduce(function (acc, [key, value]) {
-        return {...acc, [key]: value};
-      }, {});
+    this.serverVersions = _.reduce(versionConfiguration, (acc, value, key) => {
```
mykola-mokhnach: 👍

lib/server-builder.js
```diff
@@ -32,17 +33,14 @@ class ServerBuilder {
...
+      if (!VERSION_KEYS.includes(key)) {
```
mykola-mokhnach: how about making it positive?

lib/server-builder.js
```diff
@@ -107,22 +100,17 @@ class ServerBuilder {
   updateDependencyLines (configurationContent, additionalAppDependencies) {
     const dependencyLines = additionalAppDependencies
       .map(function (dependency) {
-        if (/[\s']/.test(dependency)) {
-          throw new Error(`Single quotes and space characters are disallowed in additional dependencies: ${dependency}`);
+        if (/[\s'\\$]/.test(dependency)) {
+          throw new Error(`Single quotes, dollar characters and whitespace characters are disallowed in additional dependencies: ${dependency}`);
```
mykola-mokhnach: this line seems to be too long

lib/server-builder.js
```diff
@@ -107,22 +100,17 @@ class ServerBuilder {
...
-    return configurationContent.split(/\n/)
-      .map(function (line) {
-        if (line.trim() === GRADLE_DEPENDENCIES_PLACEHOLDER) {
-          return dependencyLines;
-        } else {
-          return line;
-        }
-      })
-      .join('\n');
+      .join('$1');
```
mykola-mokhnach: will it work this way?

lib/server-builder.js
```diff
@@ -107,22 +100,17 @@ class ServerBuilder {
...
+
+    return configurationContent.replace(
+      new RegExp(`(\\s*^\\s*)${_.escapeRegExp(GRADLE_DEPENDENCIES_PLACEHOLDER)}\\s*$`, 'gm'),
```
mykola-mokhnach: Cannot `(\\s*^\\s*)` be just `(^\\s*)` ?

mykola-mokhnach:
It is necessary to make the tests green though.

lib/server-builder.js
tinder-ktarasov: `(\\s*^\\s*)` will capture new line characters currently used in the file to make sure implementation lines are correctly added. I'll add comment about it.

lib/server-builder.js
tinder-ktarasov: Yes, this uses new line characters from the capture group in the pattern. I think I was wrong that I didn't add a comment here.

lib/server-builder.js
mykola-mokhnach: I mean this particular line.
You can simply split it using `+`

tinder-ktarasov:
@mykola-mokhnach it doesn't look like the failing test case can be affected by these changes. Could you confirm that the test case about pointers is passing on `master` branch and it is not flaky?

tinder-ktarasov:
@hunzai you could check if `additionalAppDependencies` part of build configuration (see top comment in this PR) fixes issues with your app. You will need `npm link` the driver from this branch and restart Appium server to do it.

mykola-mokhnach:
@tinder-ktarasov Yep, the test is flaky. Restarting the job usually helps to resolve the problem

lib/server-builder.js
```diff
@@ -0,0 +1,160 @@
...
+    });
+    let buildLastLines = [];
+
+    const logMsg = `Output from Gradle ${this.showGradleLog ? 'will' : 'will not'} be logged`;
+    log.debug(`${logMsg}. To change this, use 'showGradleLog' desired capability`);
+    gradlebuild.on('stream-line', line => {
+      // if we have an error we want to output the logs
+      // otherwise the failure is inscrutible
+      // but do not log expected errors
+      const ignoreError = IGNORED_ERRORS.some((x) => line.includes(x));
```
mykola-mokhnach: Why do we keep IGNORED_ERRORS array empty?

lib/server-builder.js
```diff
@@ -0,0 +1,160 @@
...
+
+      if (this.showGradleLog && !ignoreError) {
+        gradleLog.error(line);
```
mykola-mokhnach: are all strings in the Gradle log errors? I would rather print them at debug or info level

lib/server-builder.js
tinder-ktarasov: For possible future filtering. Removed it now.

lib/server-builder.js
tinder-ktarasov: Probably keeping STDERR lines as `error` (that would be actual errors and task names) and changing other to `info` would be better?

lib/server-builder.js
mykola-mokhnach: YAGNI ;)

lib/server-builder.js
mykola-mokhnach: `error` is still too much. We only mark critical things as error. I would mark them as warnings and normal messages as info

mykola-mokhnach:
@tinder-ktarasov It would be nice to document the feature properly. You could put the resulting document to https://github.com/appium/appium/tree/master/docs/en/writing-running-appium/android

dpgraham:
@tinder-ktarasov do you want an invite to the Appium Slack channel?

tinder-ktarasov:
@mykola-mokhnach @KazuCocoa @dpgraham, sorry for being absent. I've added docs about this feature - https://github.com/appium/appium/pull/13766

Issue #394, opened 2019-02-11
"-android viewmatcher" selector strategy


dpgraham:
We have the `-android datamtacher` selector strategy (see https://github.com/appium/appium-espresso-driver/pull/386)

We should also have a `-android viewmatcher` selector strategy so that we have all Espresso view-finding strategies covered.

o-x-y-g-e-n:
@dpgraham  I would like to take this opportunity to add the support for view matcher.

jlipps:
thanks @o-x-y-g-e-n, let us know if you need any help; it should be pretty straightforward hopefully to base your work off the datamatcher work.

o-x-y-g-e-n:
@jlipps Yes, I got a lot of reference from the DataMatcher. For View Matcher i am stuck at one place.

for OnView(<ViewMatcher>), now ViewMatcher can be a lot of nested ViewMatchers.
eg. 

for android.support.test.espresso.matcher.ViewMatchers.withClassName it takes an argument of Matcher<String>. Hence its nested Matchers.

Let me explain you with an example. 

`{
	"using":"-android viewmatcher",
	"value":"{\"name\":\"withResourceName\",\"args\":\"io.appium.android.apis:id/text\",\"class\":\"androidx.test.espresso.matcher.ViewMatchers\"}"
}`
The above code concludes to `onView(withResourceName("io.appium.android.apis:id/text"))`

But what if i want to do something like `onView(ViewMatchers.withClassName(is("android.widget.TextView")))` which is nested by an espresso Matcher & Hamcrest Matcher.

is there any support for resolving nested Matchers in our code? If not, let me know if we have any plans to implement it.




I hope I am able to explain the problem properly.

Thanks.

KazuCocoa:
How about like backdoor? (rough idea though)

https://github.com/appium/appium-espresso-driver/blob/db2eb34c7c2ce5a05e93cca129f170941919d75e/lib/commands/general.js#L181-L200

so, for example...:

```
value: [
    {name:'withClassName'
     args: [
          {name: 'is', args: 'android.widget.TextView'}
     ]
    },
]
```

o-x-y-g-e-n:
@jlipps @dpgraham @KazuCocoa   Can we make something like `"using":"-android viewmatcher",
[...]

PR #516, opened 2019-12-06
#394 Add support for viewmatcher

o-x-y-g-e-n:
Added support for viewmatcher which almost works similar to datamatcher.


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ViewMatcherJson.kt
```diff
@@ -0,0 +1,51 @@
+package io.appium.espressoserver.lib.model
```
mykola-mokhnach: this file almost fully copies `DataMatcherJsonDeserializer`. Can it be refactor to avoid the unnecessary duplication?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.kt
```diff
@@ -189,6 +191,14 @@ object ViewFinder {
                     emptyList()
                 }
             }
+            Strategy.VIEWMATCHER -> {
+                val matcherJson = ViewMatcherJson.fromJson(selector)
+                views = try{
```
mykola-mokhnach: Please apply formatting to the file (we usually put space after try/before catch)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.kt
```diff
@@ -240,6 +250,14 @@ object ViewFinder {
         return listOf(ViewGetter().getView(dataInteractionCopy))
     }
 
+    private fun getViewsFromViewMatcher(root: View?, matcher: Matcher<View>):List<View>{
+        val viewInteraction = if(root == null)
```
mykola-mokhnach: Please format

test/functional/commands/find-e2e-specs.js
```diff
@@ -141,4 +141,85 @@ describe('find elements', function () {
...
+    it('should find an element using view matcher', async function () {
+      await driver.elementByXPath("//*[@text='Content']").click();
+      await driver.elementByXPath("//*[@text='Storage']").click();
+      await driver.elementByXPath("//*[@text='External Storage']").click();
+      await driver.element('-android viewmatcher', JSON.stringify({
+        name: 'withText',
+        args: 'Picture getExternalFilesDir',
+        class: 'androidx.test.espresso.matcher.ViewMatchers'
+      })).should.eventually.exist;
+      await driver.back();
```
mykola-mokhnach: please do not make the test dependent on each other. failure of the first test will make all the following to fail

KazuCocoa:
cc @dpgraham about viewmatcher

mykola-mokhnach:
👍 I like the deduplication refactoring 

The last questions are these dependent tests. I particularly don't like the fact, that we perform `driver.back` in several tests in order to setup the view for the following tests. This might break very easily if one of such tests fails. As a possible workaround we could put such calls into `finally` block.

What do you think @dpgraham ?

o-x-y-g-e-n:
> 👍 I like the deduplication refactoring
> 
> The last questions are these dependent tests. I particularly don't like the fact, that we perform `driver.back` in several tests in order to setup the view for the following tests. This might break very easily if one of such tests fails. As a possible workaround we could put such calls into `finally` block.
> 
> What do you think @dpgraham ?

I am glad you asked it. I actually tried to simulate the scenario where you fail the 1st test case and all the cases immediately after them actually fails. Then, i just followed the way it's done in other test cases. 

I would be more than happy to change it and totally isolate them.

mykola-mokhnach:
> Add beforeEach hook and start a new session everytime. (this will increase the runtime)

Not all of these tests require additional setup, so I would rather extract these, that work in the same view to a separate sub-suite and reinit the app for all such sub-suites to avoid calling `driver.back` in any of them in order to setup the env for following tests

o-x-y-g-e-n:
> > Add beforeEach hook and start a new session everytime. (this will increase the runtime)
> 
> Not all of these tests require additional setup, so I would rather extract these, that work in the same view to a separate sub-suite and reinit the app for all such sub-suites to avoid calling `driver.back` in any of them in order to setup the env for following tests

What i understand from this is let's divide our test cases into two different files. 
Supposedly,

1. find-e2e-specs-default.js -> This will contain all the test-cases for which driver initialization is required only once. (in Before hook).
2. find-e2e-specs.js -> This will contain all the test cases or suites for which driver initialization is required for all the test cases (in BeforeEach).

Is my understanding correct?

mykola-mokhnach:
> What i understand from this is let's divide our test cases into two different files.

not necessarily. We can still keep the stuff in the same file. The structure might be as following:

describe('independent tests') {
beforeEach() {
  startdriver()
  setupView()
}
afterEach() {
  quitDriver()
}
}

describe('dependent tests') {
before() {
  startdriver()
  setupView()
}
after() {
  quitDriver()
}
}

o-x-y-g-e-n:
> > What i understand from this is let's divide our test cases into two different files.
> 
> not necessarily. We can still keep the stuff in the same file. The structure might be as following:
> 
> describe('independent tests') {
> beforeEach() {
> startdriver()
> }
> afterEach() {
> quitDriver()
> }
> 
> ...
> 
> }
> 
> describe('dependent tests') {
> before() {
> startdriver()
> setupView()
> }
> after() {
> quitDriver()
> }
> 
> ...
> 
> }

I really understand the need of creating a new driver instance every time.  I actually tried to write some pseudo code for both logic
1. Putting driver.back() into finally block
- The limitation i faced here is that each and every time we do some assertion, that particular statement has to be under try/catch block. Now this will make our code nasty and unpredictable as there can be multiple nested try/catch block.
2. The approach you mentioned above. But the time for the complete execution increases by a very significant amount.


Conclusion i came upon,
Instead of initializing the driver every time for all the independent test cases, what we can do is we can just reset the app every time (for all the test cases). Now by doing that we have surety that no two test cases are dependent on each other and the app starts from the fresh state. i.e default activity & also I don't need to modify any test cases that are already existing. 
 
`describe.only('find elements', function () {
  this.timeout(MOCHA_TIMEOUT);

  let driver;
  before(async function () {
    driver = await initSession(APIDEMO_CAPS);
  });
  after(async function () {
    await deleteSession();
  });
 // this is the game changer!
  this.beforeEach(async function(){
    await driver.resetApp();
  })
  describe('elementByXPath', function () {
    it('should find an element by it's xpath', async function () {
      let el = await driver.elementByXPath("//*[@text='Animation']");
      el.should.exist;
      await el.click();
      await driver.back();
    });
`

mykola-mokhnach:
resetApp call won't save time, since internally it does exactly the same thing as driver.quit/driver.init.
If it was UIA2 driver then I would use terminateApp/activateApp call, but in case of Espresso the terminateApp call on the app under test would also kill the espresso runner itself.

What we could try is to use [deep linking](https://appiumpro.com/editions/84) in order to switch between views (e.g. start the corresponding intent for each view). As far as I remember in the Catalog app each entry is represented by a separate [activity](https://github.com/appium/appium-uiautomator2-server/blob/809f19f34947dc5c1d52bf5467da31bb6f7b6f99/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/GestureCommandsTest.java#L62).

o-x-y-g-e-n:
> resetApp call won't save time, since internally it does exactly the same thing as driver.quit/driver.init.
> If it was UIA2 driver then I would use terminateApp/activateApp call, but in case of Espresso the terminateApp call on the app under test would also kill the espresso runner itself.
> 
> What we could try is to use [deep linking](https://appiumpro.com/editions/84) in order to switch between views (e.g. start the corresponding intent for each view). As far as I remember in the Catalog app each entry is represented by a separate [activity](https://github.com/appium/appium-uiautomator2-server/blob/809f19f34947dc5c1d52bf5467da31bb6f7b6f99/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/GestureCommandsTest.java#L62).

I don't know much about deep linking and I think the demo app we are using doesn't support it.
But yes, each entry is represented by an activity. Hence we can use something like,
`await driver.startActivity({
        appPackage: "io.appium.android.apis",
        appActivity: ".content.ExternalStorage"
      })`
I will be soon updating all the scenarios with the following code. Does that sound right approach?

mykola-mokhnach:
>  Does that sound right approach?

yes it does

test/functional/commands/find-e2e-specs.js
```diff
@@ -10,135 +10,274 @@ chai.use(chaiAsPromised);
...
-    it('should allow "class" property with fully qualified className', async function () {
-      await driver.element('-android datamatcher', JSON.stringify({
-        name: 'notARealHamcrestMatcherStrategy', args: ['title', 'A Fake Item'], class: 'org.hamcrest.Matchers',
-      })).should.eventually.be.rejectedWith(/InvalidSelector/);
+
+    describe('View Matcher - dependent tests - Set 2', function () {
+
+      before(async function () {
+        driver = await initSession(APIDEMO_CAPS);
+        driver.startActivity({
```
mykola-mokhnach: this method should be awaited

o-x-y-g-e-n: Forgot that. So Sorry.

PR #526, opened 2020-01-03
use skipServerInstallation same way as with uiautomator2-driver

rajdeepv:
Rebuilding the espresso-server using Gradle is time-consuming on slower machines and increases startup time significantly. With uiautomator2-driver there is a capability called `skipServerInstallation` which assumes that the device already has required servers. This can be usable with espresso-driver

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -356,6 +356,7 @@ class EspressoDriver extends BaseDriver {
       showGradleLog: !!this.opts.showGradleLog,
       serverLaunchTimeout: this.opts.espressoServerLaunchTimeout,
       androidInstallTimeout: this.opts.androidInstallTimeout,
+      skipServerInstallation: this.opts.skipServerInstallation,
     });
     this.proxyReqRes = this.espresso.proxyReqRes.bind(this.espresso);
   }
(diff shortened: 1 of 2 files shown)


lib/driver.js
```diff
@@ -356,6 +356,7 @@ class EspressoDriver extends BaseDriver {
       showGradleLog: !!this.opts.showGradleLog,
       serverLaunchTimeout: this.opts.espressoServerLaunchTimeout,
       androidInstallTimeout: this.opts.androidInstallTimeout,
+      skipServerInstallation: this.opts.skipServerInstallation,
```
mykola-mokhnach: this capability should also be added to desired-caps.js

mykola-mokhnach:
I would like to add this option should be used with care. Espresso code only rebuilds the server when it is necessary, which means that having this option enabled could potentially break automated tests if the Appium code and the already installed server binary are out of sync.

By default the prebuilt server binary is stored to the `tmpDir` folder defined as https://github.com/appium/appium-base-driver/blob/8624a8061f836c14e331aeace39a352d31fcb875/lib/basedriver/driver.js#L64

Providing a custom path to this folder in capabilities (by default this is the OS temp folder root) will help to preserve the precompiled server binary between server runs and/or OS reboots.

rajdeepv:
I agree that one has to be careful while using this capability and one must clearly know what they are doing. The same is applicable with some other capabilities in the context of espresso (eg. 'noSign' capability)

Providing a custom path to server apk in capabilities is no better because even this option if enabled could potentially break automated tests if the Appium code and the already installed server apk are out of sync.

PR #536, opened 2020-01-18
implement open_notifications endpoint

rajdeepv:
Implement /session/:session_id/appium/device/open_notifications

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/OpenNotifications.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/OpenNotifications.kt
@@ -1,0 +1,28 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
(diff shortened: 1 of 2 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/OpenNotifications.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package io.appium.espressoserver.lib.handlers
+
+import io.appium.espressoserver.lib.helpers.InteractionHelper.getUiDevice
+import io.appium.espressoserver.lib.model.AppiumParams
+
+class OpenNotifications : RequestHandler<AppiumParams, Void?> {
+    override fun handleInternal(params: AppiumParams): Void? {
+        getUiDevice().openNotification()
+        return null
```
KazuCocoa: We do not need to return the result?
I mean the openNotification returns true or false.
So, should we raise an error if the result was false?

rajdeepv: makes sense. I will raise a separate PR

Issue #538, opened 2020-01-29
optionalIntentArguments support for Espresso


rajdeepv:
`optionalIntentArguments` is very helpful in setting launch intent arguments for the app. But this is only supported by UiAutomator2-driver using ADB command. Starting of the app is taken care of by espresso server and adding similar support will be helpful

mykola-mokhnach:
I can try to implement such functionality, however this would require some help from your side, particularly in testing. Would it be fine for you?

rajdeepv:
@mykola-mokhnach absolutely, I will do it

mykola-mokhnach:
Could you please check if https://github.com/appium/appium-espresso-driver/pull/539 works for you. I've added the new `intentOptions` capability of type object. This dictionary can contain the following items: https://github.com/appium/appium-espresso-driver/pull/539/files#diff-1ef70d5578e4e02509556e3883ed65a8R122

rajdeepv:
It worked for me, thanks

mykola-mokhnach:
Cool
Did you try all possible options or only some of them? I would like to see an example of the client code where the session init is done

rajdeepv:
I tried only below options:
```
intentOptions:          { 'es'  => ["qalink\:hotpanel_test_id", 'CAL_141440_095548000false'],
                                    'action'     => 'android.intent.action.MAIN',
                                    'categories' => 'android.intent.category.LAUNCHER',
                                    'component'  => 'com.bumble.app/.ui.launcher.DebugLauncherActivity',
                                    'flags'      => 'FLAG_ACTIVITY_NEW_TASK' },
```

This is Appium Server Log

```
debug] [ADB] Creating ADB subprocess with args: ["-H","localhost","-P",5037,"-s","emulator-5554","shell","am","instrument","-w","-e","debug","false","io.appium.espressoserver.test/androidx.test.runner.AndroidJUnitRunner"]
[debug] [Espresso] [Instrumentation] [STDOUT]
[Espresso] Waiting for Espresso to be online...
[debug] [WD Proxy] Matched '/status' to command name 'getStatus'
[debug] [WD Proxy] Proxying [GET /status] to [GET http://localhost:55040/status] with no body
[...]

PR #539, opened 2020-01-29
feat: Add a possibility to provide intentOptions

mykola-mokhnach:
https://github.com/appium/appium-espresso-driver/issues/538

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/IntentHelpers.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/IntentHelpers.kt
@@ -20,7 +20,6 @@ import android.content.ComponentName
 import android.content.Intent
 import android.net.Uri
 import androidx.test.platform.app.InstrumentationRegistry
-import java.lang.Exception
 import kotlin.reflect.KClass
 import kotlin.reflect.full.cast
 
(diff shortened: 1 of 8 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/IntentHelpers.kt
```diff
@@ -0,0 +1,298 @@
...
+    for ((optName, optValue) in options) {
+        val handler = handlersMapping[optName]
+                ?: throw IllegalArgumentException("The option named '$optName' is not known. " +
+                        "Only the following options are supported: ${handlersMapping.keys}")
+        handler(optName, optValue)
+    }
+
+    intent.setDataAndType(data, type)
+
+    require(hasIntentInfo) { "Either intent action, data, type or category must be supplied" }
```
rajdeepv: I got a chance to test this.
1. Currently, to succeed, all ''action', 'category', 'component' and 'flags' are needed. Therefore, we should throw this error if any of these is missing
2. OR, It will be good to have default values as
'action' => 'android.intent.action.MAIN',
'category' => 'android.intent.category.LAUNCHER',
'flags' => 'FLAG_ACTIVITY_NEW_TASK'
'category' => fullyQualifiedAppActivity()
3. We can report a specific cause of failure. eg if user missed specifying FLAG_ACTIVITY_NEW_TASK we could report failure cause as
"android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?" instead of "java.lang.RuntimeException: Could not launch activity."

mykola-mokhnach: Thanks, I'll add the necessary fixes

mykola-mokhnach: I was analysing the preconditions above and for now I don't think this all is necessary to be implemented in the code. The intent options capability itself is only going to be used by advanced engineers. So it would be fine to just properly describe the feature in the documentation. 
I would be very grateful if you could write such documentation with examples.

rajdeepv: Ok, makes sense. What is the place to write documentation?

mykola-mokhnach: https://github.com/appium/appium/tree/master/docs/en/writing-running-appium/android/

Issue #541, opened 2020-02-07
[Improvement] Need ability to Pass multiple string extras to intentOptions


pr4bh4sh:
In #539 the feature of passing intent option while using expresso driver was added. While passing single option value combination works fine, when passing multiple options only first option is actually passed to the target activity.

Examples:
1. Passing `es` params with appium in espresso mode

```
intentOptions:          { 'es'         => ["option1", 'value1', "option2", "value2"],
                                              'action'     => 'android.intent.action.MAIN',
                                              'categories' => 'android.intent.category.LAUNCHER',
                                              'component'  => 'com.bumble.app/.ui.launcher.DebugLauncherActivity',
                                              'flags'      => 'FLAG_ACTIVITY_NEW_TASK' },
```

POST request to the appium
```
[debug] [WD Proxy] Matched '/session' to command name 'createSession'
[debug] [WD Proxy] Proxying [POST /session] to [POST http://localhost:53467/session] with body: {"capabilities":{"firstMatch":[{"platform":"LINUX","webStorageEnabled":false,"takesScreenshot":true,"javascriptEnabled":true,"databaseEnabled":false,"networkConnectionEnabled":true,"locationContextEnabled":false,"warnings":{},"desired":{"platformName":"Android","platformVersion":null,"deviceName":"emulator-5554","appActivity":".ui.launcher.DebugLauncherActivity","appPackage":"com.bumble.app","appWaitActivity":"*","automationName":"espresso","skipServerInstallation":true,"noSign":true,"newCommandTimeout":0,"autoGrantPermissions":true,"skipUnlock":true,"fullReset":false,"tmpDir":"/var/folders/b9/sz7k63z90g93_sfmq86bg7yw0000gp/T/appium_temp/0","intentOptions":{"es":["option1","value1","option2","value2"],"action":"android.intent.action.MAIN","categories":"android.intent.category.LAUNCHER","component":"com.bumble.app/.ui.launcher.DebugLauncherActivity","flags":"FLAG_ACTIVITY_NEW_TASK"},...
```

Data recieved by target activity
```
D DebugLauncher: {"action":"android.intent.action.MAIN","categories":["android.intent.category.LAUNCHER"],"component":{"package":"com.bumble.app","class":"com.bumble.app.ui.launcher.DebugLauncherActivity"},"flags":335544320,"extras":{"option1":"value1"}}
```

If I start the activity with adb command
```bash
[...]

PR #542, opened 2020-02-07
feat: Allow to provide multiple values as extra activity params

mykola-mokhnach:
https://github.com/appium/appium-espresso-driver/issues/541

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/IntentHelpers.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/IntentHelpers.kt
@@ -248,7 +248,7 @@ fun makeIntent(options: Map<String, Any?>): Intent {
                 requireMap(key, value)
                         .filter { it.key is String }
                         .forEach { entry ->
-                            val numbers = requireString(entry.key as String, entry.value)
+                            requireString(entry.key as String, entry.value)
                                     .split(",")
                                     .map {
                                         try {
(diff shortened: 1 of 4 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/IntentHelpers.kt
```diff
@@ -200,88 +188,129 @@ fun makeIntent(options: Map<String, Any?>): Intent {
...
+                                            it.trim().toFloat()
+                                        } catch (e: NumberFormatException) {
+                                            throw IllegalArgumentException(
+                                                    "The value of '${entry.key}' is expected " +
+                                                            "to be a comma-separated list of float numbers. " +
+                                                            "'${entry.value}' is given instead", e)
+                                        }
+                                    }
+                                    .toFloatArray()
+                            intent.putExtra(entry.key as String, numbers)
```
KazuCocoa: [nits]
Can also use `let` to reduce `val`

```kotlin
        .forEach { entry ->
            requireString(entry.key as String, entry.value)
            .split(",")
            .map {
                try {
                    it.trim().toFloat()
                } catch (e: NumberFormatException) {
                    throw IllegalArgumentException(
                            "The value of '${entry.key}' is expected " +
                                    "to be a comma-separated list of float numbers. " +
                                    "'${entry.value}' is given instead", e)
                }
            }
            .toFloatArray().let {
              intent.putExtra(entry.key as String, it)
            }      
```

PR #543, opened 2020-02-12
Make apk signing and zipAligning through build.gradle

glebplotnikovring:
Hi!

I had an issue running instrumentation test cases via `appium-espresso-driver`. Below is the actual error. 

```
.../adb -P 5037 -s emulator-5554 shell pm install /data/local/tmp/appium_cache/0fb1bb4d76e058803bd70380c0f69e18e451795c.apk
Failure [INSTALL_PARSE_FAILED_NO_CERTIFICATES: Failed to collect certificates from /data/app/vmdl1873806080.tmp/base.apk: META-INF/CERT.SF indicates /data/app/vmdl1873806080.tmp/base.apk is signed using APK Signature Scheme v2, but no such signature was found. Signature stripped?]
```

I thought that the reason could be in different build tools. I tried to have installed `28.0.3` and `29.0.3` android build tools, but error stayed the same. Then I try build `espresso-server` locally, `zipAlign` it and `sign` it with my keystore and it resolved the error. It pushed me to thinking that there is something in `appium-espresso-server`. After some research and tries I noticed sequence: build apk -> copy it -> unzip it -> replace manifest -> zip it again and sign. I played around it a little and it appears to be the reason of my error. I checked android Gradle plugin and found that this sequence could be done at building process of apk. I modified `build.gradle` and some source files and I make it work!

Please, tell me what I need to do to have this fix in further releases of `appium-espresso-driver` :)

Thank you!


rajdeepv:
@glebplotnikovring If you have a Keystore , you can use it to sign espresso server. These are the capabilities
```
useKeystore: true
keystorePath:         
keystorePassword:     
keyAlias:             
keyPassword:          

```
and do 

`forceEspressoRebuild: true`

glebplotnikovring:
It is still signing via capabilities, but within `assembleAndroidTest` task of `espressso-server` android app. Previously, signing of the `espresso-server` apk was happened after `assembleAndroidTest` Gradle command in `EspressoRunner.buildNewModServer()`. Apk built, `AndroidManifest-test.xml` was compiled, apk unzipped, new compiled `AndroidManifest.xml` replaced the old one, apk zipped again and then signed. This sequence was the reason of the error I've posted in PR description. And because of it my application couldn't be tested with Appium and espresso driver. After some research it was found, that we can make this sequence at build time of `assembleAndroidTest` (zipAlign apk with already modified `AndroidManifest.xml` and with signature). Is it what you've asked for?

mykola-mokhnach:
^ yes, this was my question. Thanks for the explanation

lib/espresso-runner.js
```diff
@@ -48,6 +47,13 @@ class EspressoRunner {
 
     this.serverLaunchTimeout = opts.serverLaunchTimeout || ESPRESSO_SERVER_LAUNCH_TIMEOUT;
     this.androidInstallTimeout = opts.androidInstallTimeout;
+
+    // Espresso Server app needs to be signed with same keyStore as appPackage
+    if (opts.useKeystore) {
+      if (opts.keystorePath && opts.keystorePassword && opts.keyAlias && opts.keyPassword) {
```
mykola-mokhnach: we don't need to nest ifs

lib/espresso-runner.js
```diff
@@ -48,6 +47,13 @@ class EspressoRunner {
...
+        this.signingConfig = buildServerSigningConfig(opts.keystorePath, opts.keystorePassword, opts.keyAlias, opts.keyPassword);
```
mykola-mokhnach: this line is too long

lib/server-builder.js
```diff
@@ -24,6 +24,16 @@ const VERSION_KEYS = [
 
 const gradleLog = logger.getLogger('Gradle');
 
+function buildServerSigningConfig (keystoreFile, keystorePassword, keyAlias, keyPassword) {
```
mykola-mokhnach: this function has too many arguments. Consider packing them into an object

lib/espresso-runner.js
```diff
@@ -141,20 +147,11 @@ class EspressoRunner {
     await mkdirp(serverPath);
     logger.debug(`Copying espresso server template from ('${TEST_SERVER_ROOT}' to '${serverPath}')`);
     await copyGradleProjectRecursively(TEST_SERVER_ROOT, serverPath);
-    await new ServerBuilder({serverPath, buildConfiguration, showGradleLog: this.showGradleLog}).build();
+    logger.debug('Bulding espresso server');
+    await new ServerBuilder({serverPath, buildConfiguration, showGradleLog: this.showGradleLog}).build(this.appPackage, this.signingConfig);
```
mykola-mokhnach: this line is too long

lib/server-builder.js
```diff
@@ -24,6 +24,16 @@ const VERSION_KEYS = [
...
+  let json = {};
```
mykola-mokhnach: this could be simplified to
```
return {
  zipAlign: true,
  ....
}
```

lib/server-builder.js
```diff
@@ -44,7 +54,7 @@ class ServerBuilder {
     this.additionalAppDependencies = buildConfiguration.additionalAppDependencies || [];
   }
 
-  async build () {
+  async build (testAppPackage, signingConfig) {
```
mykola-mokhnach: I'd rather put arguments into an object - it might be necessary to add more later

mykola-mokhnach: perhaps, it would even make sense to pass them into the instance constructor, since they are used by multiple methods

lib/server-builder.js
```diff
@@ -56,25 +66,40 @@ class ServerBuilder {
...
       .map(key => {
         const serverVersion = this.serverVersions[key];
         const gradleProperty = `appium${key.charAt(0).toUpperCase()}${key.slice(1)}`;
-        return serverVersion ? `-P${gradleProperty}=${serverVersion}` : null;
+        return this.buildGradleProperty(gradleProperty, serverVersion);
       })
       .filter(Boolean);
 
+    if (signingConfig) {
+      args.push(this.buildGradleProperty('appiumZipAlign', signingConfig.zipAlign));
```
mykola-mokhnach: the push methods accepts multiple arguments, so this could be rewritten to 

```
args.push(
  arg1,
  arg2,
  ...
)
```

lib/server-builder.js
```diff
@@ -56,25 +66,40 @@ class ServerBuilder {
...
+
+    if (testAppPackage) {
+      args.push(this.buildGradleProperty('appiumTargetPackage', testAppPackage));
+    }
     args.push('assembleAndroidTest');
 
     return {cmd, args};
   }
 
+  buildGradleProperty (key, value) {
```
mykola-mokhnach: we don't need a separate method for this helper. This could be rather an external function or a nested function (it is anyway only used in the single method)

lib/server-builder.js
mykola-mokhnach: also it should also not be a method, but rather an external helper function

espresso-server/AndroidManifest-test.xml
```diff
@@ -1,21 +0,0 @@
-<?xml version="1.0" encoding="utf-8" standalone="no"?>
```
mykola-mokhnach: 👍

espresso-server/app/build.gradle
```diff
@@ -2,10 +2,16 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 
 android {
+    def props = getProperties()
     compileSdkVersion getIntegerProperty('appiumCompileSdk', 28)
     buildToolsVersion getStringProperty('appiumBuildTools', '28.0.3')
     defaultConfig {
-        applicationId "io.appium.espressoserver"
+        // <instrumentation android:targetPackage=""/>
```
mykola-mokhnach: is the runner name not needed?

lib/server-builder.js
glebplotnikovring: At first I made it in constructor. Then I saw that in constructor we also have configuration json, which defines tool versions. These version are the actual part of configuring `ServerBuilder` object. Modifying `espresso-server`'s `build.gradle` I figured out that to test an application with espresso on Android we need 1) to specify the application package which will be under test (because without it android wouldn't know what to test) and 2) to sign `espresso-server` apk with the same key as application under test (otherwise, SecurityException will occur at runtime). So, it makes specifying package the mandatory step in order to make espresso driver working. This pushed me to think that `ServerBuilder.build()` could actually be called several times with different packages and signingConfigs, but with the same tools configuration as them are defined in Android SDK and barely could change as often as package and signingConfig. So, I made it that way. What are your thoughts? Should I make it in constructor anyway?

espresso-server/app/build.gradle
glebplotnikovring: We are specifying `testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"` below. Or is it another runner?

lib/server-builder.js
mykola-mokhnach: I would say it makes sense to put the stuff into the constructor. We can always create a new instance when necessary.

espresso-server/app/build.gradle
mykola-mokhnach: ah, then it's fine

KazuCocoa:
Could you remove https://github.com/appium/appium-espresso-driver/blob/master/package.json#L41 this line since the file will drop by this change?

lib/espresso-runner.js
```diff
@@ -48,6 +47,16 @@ class EspressoRunner {
 
     this.serverLaunchTimeout = opts.serverLaunchTimeout || ESPRESSO_SERVER_LAUNCH_TIMEOUT;
     this.androidInstallTimeout = opts.androidInstallTimeout;
+
+    // Espresso Server app needs to be signed with same keyStore as appPackage
+    if (opts.useKeystore && opts.keystorePath && opts.keystorePassword && opts.keyAlias && opts.keyPassword) {
+      this.signingConfig = buildServerSigningConfig({
```
mykola-mokhnach: Please set this property to null otherwise

lib/server-builder.js
```diff
@@ -61,15 +74,31 @@ class ServerBuilder {
 
   getCommand () {
     const cmd = system.isWindows() ? 'gradlew.bat' : './gradlew';
+    let buildProperty = function (key, value) {
```
mykola-mokhnach: const

lib/server-builder.js
```diff
@@ -61,15 +74,31 @@ class ServerBuilder {
...
       .map(key => {
         const serverVersion = this.serverVersions[key];
         const gradleProperty = `appium${key.charAt(0).toUpperCase()}${key.slice(1)}`;
-        return serverVersion ? `-P${gradleProperty}=${serverVersion}` : null;
+        return buildProperty(gradleProperty, serverVersion);
       })
       .filter(Boolean);
 
+    if (this.signingConfig) {
+      args.push(
```
mykola-mokhnach:
```
args.push(...(
['zipAlign', 'keystoreFile', 'keystorePassword', 'keyAlias', 'keyPassword']
  .map((k) => [`appium${_.capitalize(key)}`, this.signingConfig[key]])
  .map(([k, v]) => buildProperty(k, v))
));
```

mykola-mokhnach: or simply that (if we know we only want the items of `signingConfig`):

```
args.push(...(
  _.keys(this.signingConfig)
    .map((key) => [`appium${_.capitalize(key)}`, this.signingConfig[key]])
    .map(([k, v]) => buildProperty(k, v))
));
```

mykola-mokhnach: also then there will be no need for to extract the `buildProperty` helper as it could be simply put to the lambda in `map`

KazuCocoa:
appium-espresso-driver@1.22.0

glebplotnikovring:
🙇

PR #563, opened 2020-03-31
chore: use dependabot

imurchie:


.dependabot/config.yml
```diff
@@ -0,0 +1,5 @@
+version: 1
+update_configs:
```
mykola-mokhnach: should we also use it for java deps?

imurchie: I want to watch how it works in another repo that matters much less (the apidemos app). The docs are not very good on Java/Gradle.

PR #580, opened 2020-05-21
feat: Add a possibility to set app locale

mykola-mokhnach:
Currently it is only possible to set locale via Appium Settings helper, which does it in a really [hacky](https://github.com/appium/io.appium.settings/blob/master/app/src/main/java/io/appium/settings/handlers/LocaleSettingHandler.java) manner. Also there is no guarantee this hack going to work in newer Android versions.

Although, Espresso anyway uses the target context concept, which we could change locale for without private method calls. I've added the `locale` capability that allows to set locale for an initial app startup as well as `mobile: setLocale` one, which does it dynamically.

I would really appreciate it if you could help me with testing on newer Android versions @KazuCocoa


mykola-mokhnach:
Reference link https://stackoverflow.com/questions/53417037/change-locale-while-executing-espresso-test

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/LocaleHelpers.kt
```diff
@@ -0,0 +1,37 @@
...
+
+fun changeLocale(context: Context, locale: Locale) {
+    if (Build.VERSION.SDK_INT >= 26) {
+        Locale.setDefault(Locale.Category.DISPLAY, locale)
+    } else {
+        Locale.setDefault(locale)
+    }
+    val res = context.resources
+    val config = res.configuration
+    config.locales = LocaleList(locale)
```
KazuCocoa, line 34: `config.setLayoutDirection(locale)` is necessary to apply RTL style properly

lib/desired-caps.js
```diff
@@ -40,6 +40,9 @@ let espressoCapConstraints = {
   },
   activityOptions: {
     isObject: true
+  },
+  locale: {
```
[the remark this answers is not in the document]
mykola-mokhnach: ah, ok
I think it makes sense to add a new cap instead to avoid conflicts.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/LocaleHelpers.kt
mykola-mokhnach: The docstring on setLocales method:

```
    /**
     * Set the locale list. This is the preferred way for setting up the locales (instead of using
     * the direct accessor or {@link #setLocale(Locale)}). This will also set the layout direction
     * according to the first locale in the list.
     *
     * Note that the layout direction will always come from the first locale in the locale list,
     * even if the locale is not supported by the resources (the resources may only support
     * another locale further down the list which has a different direction).
     *
     * @param locales The locale list. If null, an empty LocaleList will be assigned.
     */
```

KazuCocoa: ah, okay. I see.
I also understand this is application-wide change (via the target context), so other areas such as status bar and other applications do not change.
Will check with an app which has multiple locale/layout later

lib/desired-caps.js
```diff
@@ -41,7 +41,7 @@ let espressoCapConstraints = {
   activityOptions: {
     isObject: true
   },
-  locale: {
+  appLocale: {
```
KazuCocoa, line 44: 👍

KazuCocoa:
How did you check this feature?
I downloaded Firefox app, and set it as the app under test. The device under test was Android 10 Nokia device.

But the locale (displayed language) did not change by this via mobile command even only application-wide.

KazuCocoa:
Thanks.
Yeah, it worked in application-wide after killing running activity, and launch it again.

Even we provide appLocale capability, the initial activity does not have the appLocale one.
I mean when the device is en_US and I set `appLocale` as `appLocale: {language: 'ja', country: 'JP'},`, the initial activity has the en_US's one. Once I close the activity and launch a new or the same one, the appLocale is applied.

We should apply the locale change before starting the app under test's process to apply this from the beginning. In Espresso case, I imagine applying this change in `ActivityScenario` before launching the activity under test.

mykola-mokhnach:
> Even we provide appLocale capability, the initial activity does not have the appLocale one.

Yes, this is something I was getting as well. And this is strange, because I actually set the stuff before starting the activity

mykola-mokhnach:
Added some methods to restart the activity. Perhaps, it could make it working

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/LocaleHelpers.kt
KazuCocoa: LocaleList is only for API level 24

KazuCocoa:
I have Android 6 and Android 10 devices.
https://github.com/appium/appium-espresso-driver/pull/580#discussion_r429502632 worked for Android 6 with the current logic. When I sent mobile:setAppLocal as `{language: 'fa', country: 'AF'}`, the preference was applied the locale partially (it depends on Firefox app's implementation, I believe) after backing the app to background and moving it forward with launch the app by history.

But no luck for Android 10 even I did the same steps with Android 6.

mykola-mokhnach:
thanks for your help
Let me focus better on Android 10 then.

mykola-mokhnach:
Let me know if you have any ideas regarding the code modification

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/LocaleHelpers.kt
mykola-mokhnach: thanks, will change

mykola-mokhnach:
You were right @KazuCocoa 
We never know how to restart an activity properly, since it depends on the particular app implementation. Thus I've removed the `mobile:` endpoint and only keep the capability for setting the locale on app setup. Tested on Android 10 and Android 5 (Firefox binary): both worked as expected

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ActivityHelpers.kt
```diff
@@ -17,83 +17,107 @@
...
+                            activityField.isAccessible = true
+                            return activityField.get(activityRecord) as Activity
+                        }
                     }
+                } catch (e: Exception) {
+                    // ignore
+                }
+                return null
+            }
+            for (method in listOf(method1, method2)) {
```
KazuCocoa, line 69: 👍

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/LocaleHelpers.kt
```diff
@@ -0,0 +1,46 @@
...
+fun changeLocale(context: Context, locale: Locale) {
+    val res = context.applicationContext.resources
+    val config = res.configuration
+
+    Locale.setDefault(locale)
+    if (SDK_INT >= Build.VERSION_CODES.O) {
+        Locale.setDefault(Locale.Category.DISPLAY, locale)
+    }
+
+    if (SDK_INT >= Build.VERSION_CODES.N) {
```
KazuCocoa, line 35: 👍

PR #589, opened 2020-07-14
chore: Catch only NoMatchingViewException, do not swallow AppNotIdleException

rajdeepv:
Sometimes, the find_elements take 60 seconds and returns empty results. 

This is because app is not idle and we swallow this exception and wait for 60 seconds of a timeout to finish

We must not swallow this AppNotIdleException

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ThreadHelpers.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ThreadHelpers.kt
@@ -1,0 +1,18 @@
+package io.appium.espressoserver.lib.helpers
+
+object ThreadHelpers {
+    val threadDump: String
+        get(){
+            val threadDetails = StringBuilder()
+            val activeCount = Thread.activeCount()
+            val threads = arrayOfNulls<Thread>(activeCount)
+            Thread.enumerate(threads)
+            for (thread in threads.filterNotNull()) {
+                threadDetails.append("\n\n ThreadName: ${thread!!.name}: State: ${thread.state}")
+                for (stackTraceElement in thread.stackTrace) {
+                    threadDetails.append("\n\t\t$stackTraceElement")
+                }
+            }
+            return threadDetails.toString();
+        }
+}
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ThreadHelpers.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ThreadHelpers.kt
@@ -1,18 +1,15 @@
 package io.appium.espressoserver.lib.helpers
 
-object ThreadHelpers {
-    val threadDump: String
-        get(){
-            val threadDetails = StringBuilder()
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
I tend to think this is an expected behavior. Where this 60 seconds timeout is set? Is it the implicit timeout?

rajdeepv:
I don't think. In all other cases apart from findElement, we throw this exception to the user
This 60 second is the timeout in espresso code, maybe coming from [here](https://github.com/android/android-test/blob/461ca02299807020c0b317b1805eb709d3c25147/espresso/core/java/androidx/test/espresso/IdlingPolicies.java#L37), it can not be changed by Appium user

mykola-mokhnach:
> I don't think. In all other cases apart from findElement, we throw this exception to the user
> This 60 second is the timeout in espresso code, maybe coming from here, it can not be changed by Appium user

Yes, this makes sense to me now, thanks for the explanation. Lets only check this exception now. Also, we could add the list of running threads to add more details about what exactly is blocking Espresso (and to make the exception message more helpful):
https://stackoverflow.com/questions/38856136/espresso-test-error-appnotidleexception

rajdeepv:
Oh yes, good idea. I will do it soon

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ThreadHelpers.kt
```diff
@@ -0,0 +1,18 @@
+package io.appium.espressoserver.lib.helpers
+
+object ThreadHelpers {
```
mykola-mokhnach, line 3: kotlin allows to create public functions, there is no need for a wrapping object

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ThreadHelpers.kt
```diff
@@ -0,0 +1,18 @@
...
+    val threadDump: String
+        get(){
+            val threadDetails = StringBuilder()
+            val activeCount = Thread.activeCount()
+            val threads = arrayOfNulls<Thread>(activeCount)
+            Thread.enumerate(threads)
+            for (thread in threads.filterNotNull()) {
+                threadDetails.append("\n\n ThreadName: ${thread!!.name}: State: ${thread.state}")
```
mykola-mokhnach, line 11: !! is not needed

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.kt
```diff
@@ -268,6 +270,9 @@ object ViewFinder {
                 else
                     onView(allOf(isDescendantOfA(`is`(root)), withIndex(matcher, 0)))
                 return listOf(ViewGetter().getView(viewInteraction))
+            } catch (e: AppNotIdleException){
+                throw InvalidElementStateException("The application is expected to be in idle state in order for Espresso to interact with it." +
```
mykola-mokhnach, line 274: The message template could be extracted to a constant to avoid unnecessary duplication

PR #501, opened 2019-09-20
Android 10

dpgraham:

PR #594, opened 2020-08-19
fix: Fixed SDK 25 chromedriver version to match installed Chrome

grzegorz-jarosz:
Changed chromedriver version to match installed chrome for Android SDK 25. Using this [list as a reference](https://stackoverflow.com/questions/41133391/which-chromedriver-version-is-compatible-with-which-chrome-browser-version/49618567#49618567)
This should also fix error discussed in [Android 10 PR](https://github.com/appium/appium-espresso-driver/pull/501)


grzegorz-jarosz:
I think I may need a little help with fixing this CLA, stuff here.

mykola-mokhnach:
^ I assume it it fine, since the email has been changed between commits

thanks for the fix

grzegorz-jarosz:
Yeah, I forgot to use right one. No problem, you're welcome.

PR #497, opened 2019-09-13
Exclude transitive dependencies of espresso-contrib

tinder-ktarasov:
Should fix #449

Exclude transitive dependencies of espresso-contrib when building espresso-server APK. The dependencies include many AndroidX view libraries and thus has a big chance of conflict with AndroidX libraries in AUT.
These dependencies now would come from APK of AUT, so this change should not break tests.

Kotlin dependencies are addressed in #496
Hopefully, "non test" dependencies that are still left will not conflict with the most apps:
- androidx.lifecycle:lifecycle-common:2.0.0
- com.google.code.findbugs:jsr305:2.0.1
- com.google.code.gson:gson:2.8.5
- com.squareup:javawriter:2.1.1
- javax.inject:javax.inject:1
- javax.ws.rs:jsr311-api:1.1.1
- net.sf.kxml:kxml2:2.3.0
- org.hamcrest:hamcrest-integration:1.3
- org.ccil.cowan.tagsoup:tagsoup:1.2
- org.jetbrains:annotations:13.0
- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.3.31
- org.jetbrains.kotlin:kotlin-stdlib:1.3.31
- org.jetbrains.kotlin:kotlin-stdlib-common:1.3.31
- org.jetbrains.kotlin:kotlin-reflect:1.3.31
- org.nanohttpd:nanohttpd:2.3.1
- org.nanohttpd:nanohttpd-webserver:2.3.1

**Dependency tree for this PR**
```
+--- androidx.test.espresso:espresso-contrib:3.2.0
+--- androidx.test.espresso:espresso-core:3.2.0
|    +--- androidx.test:runner:1.2.0
|    |    +--- androidx.annotation:annotation:1.0.0
|    |    +--- androidx.test:monitor:1.2.0
|    |    |    \--- androidx.annotation:annotation:1.0.0
|    |    +--- junit:junit:4.12
|    |    |    \--- org.hamcrest:hamcrest-core:1.3
|    |    \--- net.sf.kxml:kxml2:2.3.0
|    +--- androidx.test.espresso:espresso-idling-resource:3.2.0
|    +--- com.squareup:javawriter:2.1.1
|    +--- javax.inject:javax.inject:1
|    +--- org.hamcrest:hamcrest-library:1.3
|    |    \--- org.hamcrest:hamcrest-core:1.3
|    +--- org.hamcrest:hamcrest-integration:1.3
[...]

PR #506, opened 2019-10-31
Drawable and Lifecycle issues fix (#449)

uterator:
The change has been done based on #449 discussions. 
It addresses the errors with [Drawable](https://github.com/appium/appium-espresso-driver/issues/449#issuecomment-525894658) and [Lifecycle](https://github.com/appium/appium-espresso-driver/issues/449#issuecomment-548080470)

PR #596, opened 2020-08-25
fix: Exclude transitive dependencies of espresso-contrib

grzegorz-jarosz:
This PR takes forward work started by @tinder-ktarasov in his PR #497, hoping that this will fix issues in #449.

Initial submission had some failing tests (but logs are no longer accessible). 

In a short setting "transitive = false" stops gradle from loading espresso-contrib's dependendencies. However, adding it fails one of the tests with some "Navigation" class missing. After adding "com.google.android.material:material:1.2.0" this test no longer failed.

Note that android.material package is in espresson-contrib dependencies, so in fact we're removing all but one of dependent packages from build.

I'm not sure if that's the right "gradle" solution. I'm open for a discussion.

### DIFF
--- a/espresso-server/app/build.gradle
+++ b/espresso-server/app/build.gradle
@@ -89,8 +89,10 @@ dependencies {
     testImplementation "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
 
     androidTestImplementation "androidx.annotation:annotation:$annotation_version"
-    androidTestImplementation "androidx.test.espresso:espresso-contrib:$espresso_version"
-    androidTestImplementation "androidx.test.espresso:espresso-core:$espresso_version"
+    androidTestImplementation("androidx.test.espresso:espresso-contrib:$espresso_version") {
+        transitive = false
+    }
+    androidTestImplementation "com.google.android.material:material:1.2.0"
     androidTestImplementation "androidx.test.espresso:espresso-web:$espresso_version"
     androidTestImplementation "androidx.test.uiautomator:uiautomator:$uia_version"
     androidTestImplementation "androidx.test:core:$testlib_version"
--- a/espresso-server/app/build.gradle
+++ b/espresso-server/app/build.gradle
@@ -90,6 +90,8 @@ dependencies {
 
     androidTestImplementation "androidx.annotation:annotation:$annotation_version"
     androidTestImplementation("androidx.test.espresso:espresso-contrib:$espresso_version") {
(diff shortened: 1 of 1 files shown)


espresso-server/app/build.gradle
```diff
@@ -89,8 +89,10 @@ dependencies {
     testImplementation "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
 
     androidTestImplementation "androidx.annotation:annotation:$annotation_version"
-    androidTestImplementation "androidx.test.espresso:espresso-contrib:$espresso_version"
-    androidTestImplementation "androidx.test.espresso:espresso-core:$espresso_version"
+    androidTestImplementation("androidx.test.espresso:espresso-contrib:$espresso_version") {
+        transitive = false
```
mykola-mokhnach, line 93: I would say it makes sense to add a comment here or a link the actual PR with a short explanation about why transitive deps are disabled

grzegorz-jarosz:
Somehow, I knew it won't be that easy. Okay, back to the drawing board

mykola-mokhnach:
> Somehow, I knew it won't be that easy. Okay, back to the drawing board

What exactly is complicated? The PR itself is ok

grzegorz-jarosz:
Indeed, PR looks simple, but tests are failing for SDK_30 (when I run tests locally they work) and SDK_23 (haven't looked at that one yet)

mykola-mokhnach:
they were failing even before. It looks like it's partially an azure issue (the emulator is just too slow and times out unexpectedly)

mykola-mokhnach:
simply add the comment and I would be happy to merge the PR

grzegorz-jarosz:
I've added a comment with short description. 

This may not finish all the "Drawable and Lifecycle" issues as there is other PR (#506 ) that took this work further.

In one case I had to use `"additionalAppDependencies": ["com.google.android.material:material:1.2.0"]` in espressoBuildConfig to solve some issues I had in my case. If I remember correctly those dependencies go into the gradle as "implements".

I guess I'm trying to say that there is a room for improvements.

espresso-server/app/build.gradle
grzegorz-jarosz: Added in f9b162213e159ac2c510a42d8c47a1b097c4f1c9.

mykola-mokhnach:
> I guess I'm trying to say that there is a room for improvements.

There always is. Lets do it in small steps.

PR #601, opened 2020-09-13
refactor: Improve the server startup logic

mykola-mokhnach:
It looks like on some device the instrumentation process keeps silence, so it makes sense to detect the server startup using the fact it starts responding to `/status` requests like we do in UIA2. Also, added more visibility and perf starts to the server startup process.


KazuCocoa (approved):
good idea

KazuCocoa:
mm, CI is not happy 👀

Issue #599, opened 2020-09-04
Web Atom interaction is broken due to KReflectionUtils


PochtarIvan:
During debugging with @rajdeepv we found out that the below code works fine on the version - 1.23.2 but doesn't work on the latest one.
The client code we execute is below:

```

      el = driver.find_element(get_view)

      method_chain = {
          webviewElement:         el.ref,
          forceJavascriptEnabled: true,
          methodChain:            [
              { name: 'withElement', atom: { name: 'findElement', locator: { using: 'ID', value: 'buy_button' } } },
          ],
      }
      driver.execute_script('mobile: webAtoms', method_chain)
```

we get an exception:

```
[HTTP] --> POST /wd/hub/session/62efc837-4b93-49bd-bcea-1023b61fde8a/execute/sync
[HTTP] {"script":"mobile: webAtoms","args":[{"webviewElement":"83a7abf3-c6a8-4f9f-aea6-3bd2d8d25b61","forceJavascriptEnabled":true,"methodChain":[{"name":"withElement","atom":{"name":"findElement","locator":{"using":"ID","value":"buy_button"}}},{"name":"perform","atom":"webClick"}]}]}
[debug] [W3C (62efc837)] Calling AppiumDriver.execute() with args: ["mobile: webAtoms",[{"webviewElement":"83a7abf3-c6a8-4f9f-aea6-3bd2d8d25b61","forceJavascriptEnabled":true,"methodChain":[{"name":"withElement","atom":{"name":"findElement","locator":{"using":"ID","value":"buy_button"}}},{"name":"perform","atom":"webClick"}]}],"62efc837-4b93-49bd-bcea-1023b61fde8a"]
[AndroidDriver] Executing native command 'mobile: webAtoms'
[debug] [WD Proxy] Proxying [POST /appium/execute_mobile/web_atoms] to [POST http://127.0.0.1:52796/session/98101eba-e566-4bc0-b3b4-382c0c3468ac/appium/execute_mobile/web_atoms] with body: {"webviewElement":"83a7abf3-c6a8-4f9f-aea6-3bd2d8d25b61","forceJavascriptEnabled":true,"methodChain":[{"name":"withElement","atom":{"name":"findElement","locator":{"using":"ID","value":"buy_button"}}},{"name":"perform","atom":"webClick"}]}
[...]

PR #602, opened 2020-09-14
[wip] fix: webatom

KazuCocoa:
This is still wip.(This code works only some cases. The return error also haven't tweaked.)

https://github.com/appium/appium-espresso-driver/issues/599 failed by `kclass.functions`.
The same code worked without issues on our CI like this and ruby_lib_core repo.

So, would like to ensure vanilla Java reflection might avoid the error first.


KazuCocoa:
=> https://github.com/appium/appium-espresso-driver/pull/606

PR #619, opened 2020-10-14
Fix looking up for methods in parent classes

rajdeepv:
Some Backdoors got broken recently because getDeclaredMethod finds only methods of current class. changing back to getMethod to fix the issue


mykola-mokhnach (approved):

KazuCocoa:
1.38.1 has this cahnge

rajdeepv:
@KazuCocoa Just curious is there a way to force download appium-espresso-driver version 1.38.1 ?
With the latest Appium v1.19.0-rc.0, I still get appium-espresso-driver version 1.38.0 ?

KazuCocoa:
Beta can download newer versions, but RC should get updated shrinkwrap (it requires a new push)
You can manually modify your local file downloaded as 1.38.0, but probably cannot download 1.38.1 as part of npm install.

Issue #648, opened 2020-12-29
resetApp results in an Error since 1.19.x Server Version


TomBAMU:
Recently we updated to Appium 1.19.1 and the possibility to use resetApp with the espresso-driver is not working anymore.
Unfortunately the instrumentation process crashes unexpectedly. I know the fact, that espresso is running in the actual App under Test and therefore the "clean up" is a quite delicate operation. 
Nevertheless, the functionality should be there?  Or Is this behavior intended ? Or a regression?

### Environment
Appium Version: 1.19.1
Android Version: 6.1
Android Java-Client Version: 7.3.0

### Appium Server Debug Log
```
[debug] [W3C (40b33736)] Calling AppiumDriver.execute() with args: ["mobile: shell",[{"command":"appops set com.android.music SYSTEM_ALERT_WINDOW ignore"}],"40b33736-4a26-4cdc-8057-28d5fea536d3"]
[debug] [W3C (40b33736)] Encountered internal error running command: InvalidContextError: The instrumentation process has unexpectedly crashed. Check the logcat output for more details.
[debug] [W3C (40b33736)]     at EspressoDriver.executeCommand (/home/freybergertho/.npm-global/lib/node_modules/appium/node_modules/appium-espresso-driver/lib/driver.js:439:13)
[debug] [W3C (40b33736)]     at AppiumDriver.executeCommand (/home/freybergertho/.npm-global/lib/node_modules/appium/lib/appium.js:542:36)
[debug] [W3C (40b33736)]     at process._tickCallback (internal/process/next_tick.js:68:7)
[HTTP] <-- POST /wd/hub/session/40b33736-4a26-4cdc-8057-28d5fea536d3/execute/sync 400 3 ms - 610
[HTTP] 
[HTTP] --> POST /wd/hub/session/40b33736-4a26-4cdc-8057-28d5fea536d3/appium/app/reset
[HTTP] {}
[debug] [W3C (40b33736)] Calling AppiumDriver.reset() with args: ["40b33736-4a26-4cdc-8057-28d5fea536d3"]
[debug] [W3C (40b33736)] Encountered internal error running command: InvalidContextError: The instrumentation process has unexpectedly crashed. Check the logcat output for more details.
[debug] [W3C (40b33736)]     at EspressoDriver.executeCommand (/home/freybergertho/.npm-global/lib/node_modules/appium/node_modules/appium-espresso-driver/lib/driver.js:439:13)
[...]

PR #649, opened 2020-12-29
fix: raise an error for reset

KazuCocoa:
Closes https://github.com/appium/appium-espresso-driver/issues/648

We originally expected this method raised an error. So this PR keeps it.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -31,11 +31,6 @@ commands.closeApp = async function closeApp () {
   throw new errors.UnsupportedOperationError('Please quit the session in order to close the application under test');
 };
 
-// eslint-disable-next-line require-await
-commands.reset = async function reset () {
-  throw new errors.UnsupportedOperationError('Please quit the session and create a new session in order to close and launch the application under test');
-};
-
 commands.getClipboard = async function getClipboard () {
   return (await this.adb.getApiLevel() < 29)
     ? (await this.espresso.jwproxy.command('/appium/device/get_clipboard', 'POST', {}))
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -312,7 +307,11 @@ commands.startActivity = async function startActivity (appPackage, appActivity,
 };
 
 commands.reset = async function reset () {
-  await androidHelpers.resetApp(this.adb, Object.assign({}, this.opts, {fastReset: true}));
+  try {
+    await androidHelpers.resetApp(this.adb, Object.assign({}, this.opts, {fastReset: true}));
+  } catch (e) {
+    if (!(e instanceof errors.InvalidContextError)) { throw e; }
```
mykola-mokhnach, line 313: prefer using isErrorType instead from appium-base-driver

lib/commands/general.js
```diff
@@ -31,11 +31,6 @@ commands.closeApp = async function closeApp () {
   throw new errors.UnsupportedOperationError('Please quit the session in order to close the application under test');
 };
 
-// eslint-disable-next-line require-await
-commands.reset = async function reset () {
```
mykola-mokhnach, line 35: I thought reset has been disabled for espresso.

KazuCocoa: Yes, I thought so, too.
Let's keep this line since we thought the same thing (it was disabled).

lib/commands/general.js
```diff
@@ -31,6 +30,13 @@ commands.closeApp = async function closeApp () {
   throw new errors.UnsupportedOperationError('Please quit the session in order to close the application under test');
 };
 
+// eslint-disable-next-line require-await
+commands.reset = async function reset () {
+  throw new errors.UnsupportedOperationError(
+    'Please quit the session and create a new session ' +
```
mykola-mokhnach, line 36: create a new session > create a new one

mykola-mokhnach (approved):

PR #654, opened 2021-02-07
fix: Align autoLaunch capability behaviour with iOS

mykola-mokhnach:
Related to https://github.com/appium/appium-uiautomator2-driver/pull/440


KazuCocoa (approved):

lib/driver.js
```diff
@@ -290,12 +290,11 @@ class EspressoDriver extends BaseDriver {
     } else {
       logger.debug(`'skipUnlock' capability set, so skipping device unlock`);
     }
```
111mma, line 292: w

Issue #664, opened 2021-04-05
build failing in espresso driver


KazuCocoa:
`appium-espresso-driver@1.43.0` got the below error in https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=14647&view=logs&j=04254d8a-0ee4-5202-28c9-d56c2bc94eb2&t=defac201-5385-5c4b-790f-3d632b5b92ed and failing espresso build in a new session process in Appium server side while `appium-espresso-driver@1.42.0` worked in ruby_lib_core suite.

Should figure out the cause since the same error may happen user env before our next release (at least)

```
2021-03-30T00:26:22.7183740Z [0m[1000D[KERROR AppiumLibCoreTest::PathTest#test_type (205.61s)
2021-03-30T00:26:22.7286510Z Minitest::UnexpectedError:         Selenium::WebDriver::Error::SessionNotCreatedError: A new session could not be created. Details: Unable to build Espresso server - Process ended with exitcode 1 (cmd: './gradlew -PappiumTargetPackage\=io.appium.android.apis assembleAndroidTest')
2021-03-30T00:26:22.7386440Z         Gradle error message:
2021-03-30T00:26:22.7488520Z         
2021-03-30T00:26:22.7590700Z         [STDOUT] Downloading https://services.gradle.org/distributions/gradle-6.8.3-all.zip,
2021-03-30T00:26:22.7602440Z         [STDOUT] ..............10%..............20%...............30%..............40%..............50%...............60%..............70%..............80%...............90%..............100%,
2021-03-30T00:26:22.7704760Z         [STDOUT] Starting a Gradle Daemon (subsequent builds will be faster),
2021-03-30T00:26:22.7806910Z         [STDERR] ,
2021-03-30T00:26:22.7908790Z         [STDERR] FAILURE: Build failed with an exception.,
2021-03-30T00:26:22.8010760Z         [STDERR] ,
2021-03-30T00:26:22.8088210Z         [STDERR] * What went wrong:,
2021-03-30T00:26:22.8192880Z         [STDERR] Task 'assembleAndroidTest' not found in root project 'espresso-server-emulator-5554'.,
2021-03-30T00:26:22.8230450Z         [STDERR] ,
2021-03-30T00:26:22.8261530Z         [STDERR] * Try:,
[...]

PR #665, opened 2021-04-05
chore: tweak assembleAndroidTest command

KazuCocoa:
https://github.com/appium/appium-espresso-driver/issues/664 failed because of no proper command on the Azure host.
Potentially we should add `app:`  for some environment.

Let me try this change out.

(My local env had no issue with/without `app:`)

### DIFF
--- a/espresso-server/settings.gradle.kts
+++ b/espresso-server/settings.gradle.kts
@@ -1,0 +1,1 @@
+include(":app")
--- a/espresso-server/settings.gradle
+++ b/espresso-server/settings.gradle
@@ -1,1 +1,0 @@
-include ':app'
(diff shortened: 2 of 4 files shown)


mykola-mokhnach (approved):

KazuCocoa:
https://github.com/appium/ruby_lib_core/pull/323 is if this works

espresso-server/settings.gradle.kts
```diff
@@ -0,0 +1 @@
+include(":app")
```
KazuCocoa, line 1: This file was renamed to this in https://github.com/appium/appium-espresso-driver/blob/master/package.json#L39 so I guess this was expected

KazuCocoa: cc @misa

KazuCocoa:
This change seems correct.
https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=14766&view=logs&j=3cea53da-5fe4-517a-f510-8918ac74238e was green

PR #666, opened 2021-04-16
feat: add disableAnalytics true

KazuCocoa:
https://github.com/appium/appium-uiautomator2-driver/pull/452 for espresso

### DIFF
--- a/lib/espresso-runner.js
+++ b/lib/espresso-runner.js
@@ -231,6 +231,7 @@ class EspressoRunner {
       'am', 'instrument',
       '-w',
       '-e', 'debug', process.env.ESPRESSO_JAVA_DEBUG === 'true',
+      '-e', 'disableAnalytics', true, // To avoid unexpected error by google analytics
     ];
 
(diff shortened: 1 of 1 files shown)


lib/espresso-runner.js
```diff
@@ -231,6 +231,7 @@ class EspressoRunner {
       'am', 'instrument',
       '-w',
       '-e', 'debug', process.env.ESPRESSO_JAVA_DEBUG === 'true',
+      '-e', 'disableAnalytics', true, // To avoid unexpected error by google analytics
```
mykola-mokhnach, line 234: I assume true should be a string rather than a boolean

mykola-mokhnach (approved):

lib/espresso-runner.js
KazuCocoa: It seems fine with the current way.
When I ran with this line (both `true` and `'true'`), logcat got `UsageTrackerFacilitator: Usage tracking disabled`. But without this line, nothing appeared.

PR #672, opened 2021-05-21
feat: add ability for driver to recieve 'reboot' and 'suppress-kill-s…

khanayan123:
…erver' args


ayvnkhan:
depends on this https://github.com/appium/appium-android-driver/pull/693

lib/driver.js
```diff
@@ -104,11 +104,14 @@ const APP_EXTENSION = '.apk';
 
 
 class EspressoDriver extends BaseDriver {
-  constructor (opts = {}, shouldValidateCaps = true) {
+  constructor (opts = {}, shouldValidateCaps = true, driverArgs = {}) {
     // `shell` overwrites adb.shell, so remove
     delete opts.shell;
 
-    super(opts, shouldValidateCaps);
+    opts.reboot = helpers.parseReboot(driverArgs);
```
mykola-mokhnach, line 111: are these parsers really needed? I would simply pass args as is

jlipps: i agree that we don't need parsers since these values are coming in as JSON converted to JS objects. the only reason we'd ever need parsers is if the value coming in from the JSON is not the final value. for example, if it's a CSV (but in that case we could make the value an array instead!), or a filepath

jlipps (changes requested):

jlipps (commented):
i agree with @mykola-mokhnach. i think we should standardize with all these drivers on storing driver args in a map just like we store capabilities. they can even be validated by the same validation logic.

docs/android-espresso.md
```diff
@@ -0,0 +1,13 @@
...
+### Server Arguments
+
+Appium 2.0 Usage: `node . --driver-args='{"espresso": {[argName]: [argValue]}}'`
+
+<expand_table>
+
+|Argument|Default|Description|Example|
+|----|-------|-----------|-------|
+
+|`"reboot"`|false|reboot emulator after each session and kill it at the end|`--driver-args='{"espresso": {"reboot": true}}'`|
```
KazuCocoa, line 12: Wrong table format?

jlipps (commented):
looks good but CI is failing. I'm not sure if it normally passes these days...?

ayvnkhan:
> looks good but CI is failing. I'm not sure if it normally passes these days...?
Judging based off of the closed branch history, I assume not? unit tests pass, but e2e tests seem to be having issues

jlipps:
Oh in this case the error is because the validation function has not yet been published so of course it will fail

mykola-mokhnach:
Similar comments as for https://github.com/appium/appium-xcuitest-driver/pull/1307
And yes, it only makes sense to publish this module after we bump the beta version of base driver for v2

mykola-mokhnach:
closed as obsolete

PR #690, opened 2021-09-20
feat: add compose handler to find multiple elements

rajdeepv:
Add compose handler to find multiple elements

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/drivers/ComposeDriver.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/drivers/ComposeDriver.kt
@@ -29,7 +29,7 @@ class ComposeDriver : AppDriver {
     override val name = DriverContext.StrategyType.COMPOSE
 
     override fun findElement(params: Locator): BaseElement {
-        val nodeInteractions = nodeInteractionCollection(params)
+        val nodeInteractions = toNodeInteractionsCollection(params)
         if (nodeInteractions.fetchSemanticsNodes(false).isEmpty()) throw NoSuchElementException(
             String.format(
                 "Could not find a compose element with strategy '%s' and selector '%s'",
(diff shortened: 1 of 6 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/drivers/ComposeDriver.kt
```diff
@@ -29,22 +29,31 @@ class ComposeDriver : AppDriver {
...
+                params.using, params.value
             )
-        }
+        )
         return ComposeElement(nodeInteractions[0])
     }
 
+    override fun findElements(params: Locator): List<BaseElement> {
+        val nodeInteractions = nodeInteractionCollection(params)
+        val elements = mutableListOf<BaseElement>()
```
mykola-mokhnach, line 44: `return nodeInteractions.fetchSemanticsNodes(false). map { ComposeElement(nodeInteractions[it]) }`

rajdeepv: `it` inside is not integer index, but `SemanticsNode`. 
Something like this could work:
```
override fun findElements(params: Locator): List<BaseElement> {
        val nodeInteractions = toNodeInteractionsCollection(params)
        return nodeInteractions.fetchSemanticsNodes(false)
            .mapIndexed { index, semanticsNode -> ComposeElement(nodeInteractions[index]) }
    }
```

mykola-mokhnach: Works for me

rajdeepv: changed

mykola-mokhnach (approved):

PR #691, opened 2021-09-22
chore: Switch to Kotlin 1.5 and address deprecations

mykola-mokhnach:
Compose driver anyway requires Kotlin 1.5, so we get warnings about two kotlin version being used at the same time during compilation. Also fixed several compiler warnings related to the above version bump.

cc @rajdeepv


KazuCocoa (approved):
kotlin & code changes in kt looks good

> CI
mm

Issue #684, opened 2021-08-09
jetpack compose support


rajdeepv:
<< ref: https://github.com/appium/appium/issues/15138 >>

We are migrating our app to compose and plan to add composeTestRule support inside appium-espresso-driver. Before moving forward with any change, I wanted to check upon the approach so that it could be integrated into existing code seamlessly.
We plan to introduce a switch between espresso and composeTestRule. we could utilize switch_context or execute_script to switch between contexts. on the espresso-server side, based on the current context of the driver, we can use either implementation. Please suggest if you have any other ideas

mykola-mokhnach:
context switch seems promising for such purpose, although I'm not sure it would be easy to update its current implemntation without breaking many things as it was only created to switch to Chrome web views. What might be simpler is to organize context switch using settings API

rajdeepv:
Haven't seen espresso using settings API, any pointer to it will be helpful. we could also do something like execute_script("mobile: switch_to_compose") or execute_script("mobile: switch_to_espresso") ?

mykola-mokhnach:
yes. Something like `mobile: setServerContext` (name: `classic` or `compose`) and `mobile: getServerContext`

Examples: https://github.com/appium/appium-mac2-driver/pull/28 https://github.com/appium/appium-mac2-driver/pull/30

jlipps:
it would still be worth exploring the strategy of using the official context API. if that works sensibly, it will be a much better user experience.

mykola-mokhnach:
I remember I was trying to make the context stuff working with Firefox. My findings were:
- Most of the code lives in android driver. Changing that will affect all Android drivers
- The current implementation is very Chrome-specific. It was not developed for further extension
- Some parts of the codebase a quite cryptic. It might take a lot of time to understand them not even talking about how to change them without breaking the existing functionality

jlipps:
You're probably right. I was thinking of something simple like a wrapper around the existing functionality that simply adds a new context to the list and handles the case of selecting that context. Otherwise, everything is handled by just calling the existing methods.

KazuCocoa:
Its promising!
Initially I thought a bit:

- define a context like WebView/Native to be able to switch to `compose`
- define a new look up like `-android viewmatcher` to manage an element object
    - add `mobile:` for new actions if needed
- add `mobile: getComposSource` https://developer.android.com/jetpack/compose/testing#debugging as a workaround
    - until we add xml converter

[...]

PR #692, opened 2021-09-22
feat: support find by xpath, and page_source

rajdeepv:
- Support finding elements by Xpath
- Support getting page source


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,93 @@
...
+                bounds.left.toInt(), bounds.top.toInt(), bounds.right.toInt(),
+                bounds.bottom.toInt()
+            )
+        }
+
+    val resourceId: String
+        get() = node.id.toString()
+
+    val text: String
+        get() = node.config.getOrNull(SemanticsProperties.Text)?.get(0).toString()
```
mykola-mokhnach, line 46: we also need to verify the array length

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    val index: Int
+        get() {
+            val parent = node.parent
```
mykola-mokhnach, line 50: do we need this intermediate variable?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,93 @@
...
+                true
+            } catch (e: AssertionError) {
+                false
+            }
+        }
+
+    private fun getSemanticsNodeInteraction() =
+        EspressoServerRunnerTest.composeTestRule.onNode(hasId(node.id))
+
+    private fun hasId(id: Int): SemanticsMatcher {
```
mykola-mokhnach, line 85: this could be a one-liner

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -199,9 +206,49 @@ class SourceDocument constructor(
         serializer?.endTag(NAMESPACE, tagName)
     }
 
+    private fun serializeComposeNode(semanticsNode: SemanticsNode?, depth: Int) {
+        if (semanticsNode == null) {
+            return
+        }
+        val nodeElement = ComposeNodeElement(semanticsNode)
+        serializer?.startTag(NAMESPACE, COMPOSE_TAG_NAME)
```
mykola-mokhnach, line 214: cannot we retrieve the class name of the compose node?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -219,7 +266,17 @@ class SourceDocument constructor(
                         it.startDocument(XML_ENCODING, true)
                         it.setFeature("http://xmlpull.org/v1/doc/features.html#indent-output", true)
                         val startTime = SystemClock.uptimeMillis()
-                        serializeView(rootView, 0)
+                        when(context.driverStrategy.name){
```
mykola-mokhnach, line 269: please format

mykola-mokhnach:
Could you please provide an example of source XML?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,93 @@
...
+import android.graphics.Rect
+import androidx.compose.ui.semantics.*
+import androidx.compose.ui.test.*
+import io.appium.espressoserver.EspressoServerRunnerTest
+import java.lang.AssertionError
+
+class ComposeNodeElement(private val node: SemanticsNode) {
+
+    val contentDescription: CharSequence?
+        get() = node.config.getOrNull(SemanticsProperties.ContentDescription)?.get(0)
```
mykola-mokhnach, line 28: we should verify the array length here

rajdeepv:
> Could you please provide an example of source XML?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
rajdeepv: I tried, looks like they don't have any specific class as such

mykola-mokhnach: I assume we could use https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/Role for this purpose

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -199,9 +206,49 @@ class SourceDocument constructor(
...
 
+    private fun serializeComposeNode(semanticsNode: SemanticsNode?, depth: Int) {
+        if (semanticsNode == null) {
+            return
+        }
+        val nodeElement = ComposeNodeElement(semanticsNode)
+        serializer?.startTag(NAMESPACE, COMPOSE_TAG_NAME)
+
+        linkedMapOf(
+            ViewAttributesEnum.INDEX to { nodeElement.index },
```
mykola-mokhnach, line 217: I would say it also makes sense to expose https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/SemanticsProperties#Password(), https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/SemanticsProperties#ProgressBarRangeInfo() and https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/SemanticsProperties#InvisibleToUser()

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,93 @@
...
+    val isScrollable: Boolean
+        get() = node.config.contains(SemanticsActions.ScrollBy)
+
+    val isSelected: Boolean
+        get() = node.config.getOrNull(SemanticsProperties.Selected) == true
+
+    val isVisible: Boolean
+        get() {
+            return try {
+                getSemanticsNodeInteraction().assertIsDisplayed()
```
mykola-mokhnach, line 75: this might be expensive as we'd have to request the UI for each node. Could we use https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/SemanticsProperties#InvisibleToUser() instead?

mykola-mokhnach:
do we need to expose viewIndex in the XML? I can observe it is always the same as resourceId

mykola-mokhnach:
I don't think we show text as `null` in Espresso id if it does not present. It is rather transformed to an empty string

mykola-mokhnach:
Could you please rebase this PR with master, so we check how it works together?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
rajdeepv: Most of the time `SemanticsProperties.Role` is null. What should we do in such cases?

mykola-mokhnach: Use the default value like we do now (e.g. `ComposeNode`)

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
mykola-mokhnach: I was thinking about we could have a `firstObject` extension method  on a List class. This method should be safe to call (e.g.) return null if the list has no elements or the actual value of the first element and will allow us to reduce the amount of the boilerplate code

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: Build fails with `This API is experimental and is likely to change in the future.`

mykola-mokhnach: then probably don't include it to xml yet to avoid unnecessary slowness while querying all elements?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: I found there is already such a method https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first-or-null.html

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
rajdeepv: does that mean some nodes will be having Role as a class attribute and some of them will have no class attribute?

rajdeepv: I think `SemanticsProperties.Role` is more of an attribute of the node than the node tag name itself. 
Anyway, I am open to change it to anything you suggest

mykola-mokhnach: For there that have role null simply set the default ComposeNode value. This is the same we do in UIA2, just setting it to android.view.View instead

rajdeepv:
@mykola-mokhnach Now, xml would look like this
```
<?xml version='1.0' encoding='UTF-8' standalone='yes' ?>
<ComposeNode class="ComposeNode" index="0" clickable="false" enabled="true" focused="false" scrollable="false"
             selected="false" bounds="[0,0][1080,1962]" resource-id="37">
  <ComposeNode class="ComposeNode" index="0" clickable="false" enabled="true" focused="false" scrollable="true"
               selected="false" bounds="[0,0][1080,1962]" resource-id="38">
    <ComposeNode class="ComposeNode" index="0" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[44,0][1036,40]" text="This is a discrete slider" resource-id="83"/>
    <ComposeNode class="ComposeNode" index="1" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[22,106][1058,238]" resource-id="85">
      <ComposeNode class="ComposeNode" index="0" clickable="false" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[22,106][1058,238]"
                   progress="ProgressBarRangeInfo(current=0.0, range=0.0..10.0, steps=4)" resource-id="86"/>
    </ComposeNode>
    <ComposeNode class="ComposeNode" index="2" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[22,282][314,334]" text="Slider value is 0.0" resource-id="89"/>
    <ComposeNode class="ComposeNode" index="3" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[44,400][1036,497]" text="This is a checkbox that represents two states"
                 resource-id="91"/>
    <ComposeNode class="ComposeNode" index="4" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[22,563][1058,718]" resource-id="93">
      <Checkbox class="Checkbox" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
                selected="false" bounds="[66,607][133,674]" resource-id="94"/>
      <ComposeNode class="ComposeNode" index="1" clickable="false" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[155,607][535,659]" text="Use Jetpack Compose" resource-id="95"/>
    </ComposeNode>
    <ComposeNode class="ComposeNode" index="5" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[44,784][1036,881]" text="This is a checkbox that represents three states"
                 resource-id="97"/>
    <ComposeNode class="ComposeNode" index="6" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[22,947][1058,1102]" resource-id="99">
      <Checkbox class="Checkbox" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
                selected="false" bounds="[66,991][133,1058]" resource-id="100"/>
      <ComposeNode class="ComposeNode" index="1" clickable="false" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[155,991][535,1043]" text="Use Jetpack Compose" resource-id="101"/>
    </ComposeNode>
    <ComposeNode class="ComposeNode" index="7" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[44,1168][1036,1220]" text="This is a radio button group" resource-id="103"/>
    <ComposeNode class="ComposeNode" index="8" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[22,1286][1058,1487]" resource-id="105">
      <ComposeNode class="ComposeNode" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
                   selected="true" bounds="[22,1286][1058,1353]" text="Android" resource-id="106">
        <RadioButton class="RadioButton" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
                     selected="true" bounds="[66,1286][133,1353]" resource-id="108"/>
      </ComposeNode>
      <ComposeNode class="ComposeNode" index="1" clickable="true" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[22,1353][1058,1420]" text="iOS" resource-id="112">
        <RadioButton class="RadioButton" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
                     selected="false" bounds="[66,1353][133,1420]" resource-id="114"/>
      </ComposeNode>
      <ComposeNode class="ComposeNode" index="2" clickable="true" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[22,1420][1058,1487]" text="Windows" resource-id="118">
        <RadioButton class="RadioButton" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
                     selected="false" bounds="[66,1420][133,1487]" resource-id="120"/>
      </ComposeNode>
    </ComposeNode>
    <ComposeNode class="ComposeNode" index="9" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[44,1553][1036,1605]" text="This is a switch component" resource-id="124"/>
    <ComposeNode class="ComposeNode" index="10" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[22,1671][1058,1826]" resource-id="126">
      <Switch class="Switch" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
              selected="false" bounds="[66,1715][172,1782]" resource-id="127"/>
      <ComposeNode class="ComposeNode" index="1" clickable="false" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[194,1715][500,1767]" text="Enable Dark Mode" resource-id="128"/>
    </ComposeNode>
    <ComposeNode class="ComposeNode" index="11" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[44,1892][1036,1962]" text="This is how you add a ripple effect to a view"
                 resource-id="130"/>
    <ComposeNode class="ComposeNode" index="12" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[0,0][1036,132]" resource-id="132">
      <ComposeNode class="ComposeNode" index="0" clickable="true" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[0,0][1036,132]" text="Click Me" resource-id="133"/>
    </ComposeNode>
    <ComposeNode class="ComposeNode" index="13" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[0,0][0,0]" text="This is a non-discrete slider" resource-id="77"/>
    <ComposeNode class="ComposeNode" index="14" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[0,0][0,0]" resource-id="79">
      <ComposeNode class="ComposeNode" index="0" clickable="false" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[0,0][0,0]"
                   progress="ProgressBarRangeInfo(current=0.0, range=0.0..1.0, steps=0)" resource-id="80"/>
    </ComposeNode>
    <ComposeNode class="ComposeNode" index="15" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[0,0][0,0]" text="This is a material Snackbar" resource-id="69"/>
    <ComposeNode class="ComposeNode" index="16" clickable="false" enabled="true" focused="false" scrollable="false"
                 selected="false" bounds="[0,0][0,0]" resource-id="71">
      <ComposeNode class="ComposeNode" index="0" clickable="false" enabled="true" focused="false" scrollable="false"
                   selected="false" bounds="[0,0][0,0]" resource-id="72">
        <ComposeNode class="ComposeNode" index="0" clickable="false" enabled="true" focused="false" scrollable="false"
                     selected="false" bounds="[0,0][0,0]" text="I'm a very nice Snackbar" resource-id="73"/>
        <ComposeNode class="ComposeNode" index="1" clickable="false" enabled="true" focused="false" scrollable="false"
                     selected="false" bounds="[0,0][0,0]" text="OK" resource-id="75"/>
      </ComposeNode>
    </ComposeNode>
  </ComposeNode>
</ComposeNode>
```

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,67 @@
...
+
+import android.graphics.Rect
+import androidx.compose.ui.semantics.*
+
+class ComposeNodeElement(private val node: SemanticsNode) {
+
+    val contentDescription: CharSequence? =
+        node.config.getOrNull(SemanticsProperties.ContentDescription)?.firstOrNull()
+
+    val text: CharSequence? = node.config.getOrNull(SemanticsProperties.Text)?.firstOrNull()
```
mykola-mokhnach, line 27: null values should be translated to an empty text for the xml

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,67 @@
...
+
+package io.appium.espressoserver.lib.model
+
+import android.graphics.Rect
+import androidx.compose.ui.semantics.*
+
+class ComposeNodeElement(private val node: SemanticsNode) {
+
+    val contentDescription: CharSequence? =
+        node.config.getOrNull(SemanticsProperties.ContentDescription)?.firstOrNull()
```
mykola-mokhnach, line 25: null values should be translated to an empty text for the xml representation

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    val text: CharSequence? = node.config.getOrNull(SemanticsProperties.Text)?.firstOrNull()
+
+    val resourceId: String = node.id.toString()
+
+    val viewTag: CharSequence? = node.config.getOrNull(SemanticsProperties.TestTag)
```
mykola-mokhnach, line 31: same as above

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,67 @@
...
+    val isEnabled: Boolean = !node.config.contains(SemanticsProperties.Disabled)
+
+    val isFocused: Boolean = node.config.getOrNull(SemanticsProperties.Focused) == true
+
+    val isScrollable: Boolean = node.config.contains(SemanticsActions.ScrollBy)
+
+    val isSelected: Boolean = node.config.getOrNull(SemanticsProperties.Selected) == true
+
+    val className: String =
+        node.config.getOrNull(SemanticsProperties.Role)?.toString() ?: "ComposeNode"
```
mykola-mokhnach, line 44: Please move `ComposeNode` to constants

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    val progress: String? =
+        node.config.getOrNull(SemanticsProperties.ProgressBarRangeInfo)?.toString()
```
mykola-mokhnach, line 47: if an element has range info then we should assign its current value to `text` property. min and max value should also be represented as separate min/max attributes if exist

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,67 @@
...
+    val bounds: Rect
+        get() {
+            val bounds = node.boundsInWindow
+            return Rect(
+                bounds.left.toInt(), bounds.top.toInt(), bounds.right.toInt(),
+                bounds.bottom.toInt()
+            )
+        }
+
+    val isPassword: Boolean? = node.config.getOrNull(SemanticsProperties.Password)?.let { true }
```
mykola-mokhnach, line 66: it should be either true or false. No need to expose null

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
mykola-mokhnach: I'd say for now lets only add text, because the other don't also exist in espresso

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -278,9 +350,37 @@ class SourceDocument constructor(
...
+                                VIEW_INDEX
+                            )
+                        )
+                    )
                 }
             }
         }, { performCleanup() })
     }
+
+    fun hasXpath(xpathSelector: String): SemanticsMatcher {
```
mykola-mokhnach, line 365: private?

mykola-mokhnach:
I've recently spend some time to stabilize e2e tests in master. Feel free to rebase one more time to make them pass

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: In Espresso, we do not add `content-desc` or `text` in xml if the values are null.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: same as above

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: For espresso, `view-tag` is also not present in the XML if it is null.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
rajdeepv: It is being used in another file

mykola-mokhnach: I'm not quite sure this method logically belongs to SourceDocument entity. Maybe it makes sense to extract it to a separate module?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
mykola-mokhnach: yeah, I would say we should reconsider that. At some point of time. Basically my reason is that looking for an empty attribute is simpler and more straighforward in xpath queries in comparison to `no-attribute` queries, e.g.:
`//*[not(@text)]` vs `//*[@text=""]`

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
rajdeepv: I can move it to `ComopseNodeFinder.kt`. This will demand some methods of `SourceDocument ` to be made public/internal

rajdeepv: I kept it there because there was counterpart method `SourceDocument#findViewsByXPath` for espresso

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: I was thinking if there is any disadvantage of having lots of empty attributes, can it affect XPath lookup speed while matching specific text values?

mykola-mokhnach: I assume one won't see the difference in perf unless the page source itself would have hundreds on megabytes in size which will never be the case for what we are doing. Having empty attributes is more preferable from a user perspective, because it is always visible which attribute names could be present there

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -76,11 +81,11 @@ private fun toXmlNodeName(className: String?): String {
...
 
 class SourceDocument constructor(
-        private val root: View? = null,
-        private val includedAttributes: Set<ViewAttributesEnum>? = null
+    private val root: Any? = null,
+    private val includedAttributes: Set<ViewAttributesEnum>? = null
 ) {
     @Suppress("PrivatePropertyName")
-    private val RESOURCES_GUARD = Semaphore(1)
+    val RESOURCES_GUARD = Semaphore(1)
```
mykola-mokhnach, line 88: why not private?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,72 @@
...
+    val resourceId: String = node.id.toString()
+
+    val viewTag: CharSequence = node.config.getOrElse(SemanticsProperties.TestTag) { "" }
+
+    val isClickable: Boolean = node.config.contains(SemanticsActions.OnClick)
+
+    val isEnabled: Boolean = !node.config.contains(SemanticsProperties.Disabled)
+
+    val isFocused: Boolean =
+        node.config.getOrNull(SemanticsProperties.Focused)?.let { true } ?: false
```
mykola-mokhnach, line 43: `== true` was also ok

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ComposeNodeFinder.kt
```diff
@@ -57,3 +64,28 @@ fun getSemanticsNode(elementId: String): SemanticsNode =
...
+    val sourceDocument = SourceDocument(parentNode, fetchIncludedAttributes(xpathSelector))
+    val expr = try {
+        XPATH.compile(xpathSelector)
+    } catch (xe: XPathExpressionException) {
+        throw XPathLookupException(xpathSelector, xe.message!!)
+    }
+    return SemanticsMatcher(
+        "Matches Xpath $xpathSelector"
+    ) {
+        val nodeIndices = sourceDocument.RESOURCES_GUARD.withPermit({
```
mykola-mokhnach, line 80: I don't think we should use the same resources guard here. The matcher should have a separate one

mykola-mokhnach: perhaps we should just hide this internal implementation details in SourceDocument class and only pass the matcher function there as an argument

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,72 @@
...
+package io.appium.espressoserver.lib.model
+
+import android.graphics.Rect
+import androidx.compose.ui.semantics.*
+
+const val COMPOSE_TAG_NAME = "ComposeNode"
+
+class ComposeNodeElement(private val node: SemanticsNode) {
+
+    val contentDescription: CharSequence? =
```
mykola-mokhnach, line 26: I'm not quite sure we need to transform null to en empty string here. The reason for it is when we implement getting attributes then it does name sense to returns `nulls` there. So I would rather perform this transformation inside of XML builder

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ComposeNodeFinder.kt
rajdeepv: I am not clear what you mean, can you help with an example?

rajdeepv: Do you mean something like this in SourceDocument
```
fun <R> hasXpath(xpathSelector: String, block: (List<Int>) -> R): R {
        val expr = try {
            XPATH.compile(xpathSelector)
        } catch (xe: XPathExpressionException) {
            throw XPathLookupException(xpathSelector, xe.message!!)
        }
        return block(nodeIndices(expr))
    }

    private fun nodeIndices(expr: XPathExpression): List<Int> {
        val nodeIndices = RESOURCES_GUARD.withPermit({
            toStream().use { xmlStream ->
                val list =
                    expr.evaluate(InputSource(xmlStream), XPathConstants.NODESET) as NodeList
                (0 until list.length).map { index ->
                    list.item(index).attributes.getNamedItem(ViewAttributesEnum.RESOURCE_ID.toString()).nodeValue.toInt()
                }
            }
        }, { performCleanup() })
        return nodeIndices
    }
```
And we invoke it from ComposeNodeFinder like below
```
private fun hasXpath(locator: Locator): SemanticsMatcher {
    return SourceDocument(
        locator.elementId?.let { getSemanticsNode(it) },
        fetchIncludedAttributes(locator.value!!)
    ).hasXpath(locator.value!!) { nodeIndices: List<Int> ->
        SemanticsMatcher("Matches Xpath ${locator.value}") { nodeIndices.contains(it.id) }
    }
}
```

mykola-mokhnach: almost. hasXpath could stay where it is now and only 

```
val nodeIndices = RESOURCES_GUARD.withPermit({
            toStream().use { xmlStream ->
                val list =
                    expr.evaluate(InputSource(xmlStream), XPathConstants.NODESET) as NodeList
                (0 until list.length).map { index ->
                    list.item(index).attributes.getNamedItem(ViewAttributesEnum.RESOURCE_ID.toString()).nodeValue.toInt()
                }
            }
        }, { performCleanup() })
```

part, which would return matched node indexes and accepts the xpath expression would belong to the SourceDocument

mykola-mokhnach: let me know if you are still unsure on how to design it. I could add the patch directly to this PR

rajdeepv: Do you mean something like this?
```
# SourceDocuemt
fun matchingNodesIndices(xpathSelector: String): List<Int> {
    val expr = try {
        XPATH.compile(xpathSelector)
    } catch (xe: XPathExpressionException) {
        throw XPathLookupException(xpathSelector, xe.message!!)
    }
    return RESOURCES_GUARD.withPermit({
        toStream().use { xmlStream ->
            val list =
                expr.evaluate(InputSource(xmlStream), XPathConstants.NODESET) as NodeList
            (0 until list.length).map { index ->
                list.item(index).attributes.getNamedItem(ViewAttributesEnum.RESOURCE_ID.toString()).nodeValue.toInt()
            }
        }
    }, { performCleanup() })
}

# ComposeNodeFinder.kt
private fun hasXpath(locator: Locator): SemanticsMatcher {
    val sourceDocument = SourceDocument(
        locator.elementId?.let { getSemanticsNode(it) },
        fetchIncludedAttributes(locator.value!!)
    )
    return SemanticsMatcher("Matches Xpath ${locator.value}") {
        sourceDocument.matchingNodesIndices(
            locator.value!!
        ).contains(it.id)
    }
}
```

mykola-mokhnach: yes, this makes sense. Also consider unifying `getMatchingNodeIds` method, so it could also be reused for espresso nodes lookup and to avoid extra code duplication if possible

KazuCocoa:
link: https://github.com/appium/appium-espresso-driver/issues/684

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/XMLHelpers.kt
```diff
@@ -21,36 +21,43 @@ import java.util.regex.Pattern
...
+            .matcher(str)
+            .replaceAll("")
         return XML10_TAG_PATTERN
-                .matcher(nodeName)
-                .replaceAll("_")
+            .matcher(nodeName)
+            .replaceAll("_")
     }
 
     fun toSafeString(source: Any?, replacement: String): String? {
```
mykola-mokhnach, line 55: in such case the result is not nullable

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/XMLHelpers.kt
```diff
@@ -21,36 +21,43 @@ import java.util.regex.Pattern
...
         return if (source == null)
-            null
+            ""
```
mykola-mokhnach, line 57: this change might break existing espresso stuff. Please verify XML files generated by Espresso did not change

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -0,0 +1,71 @@
...
+    val isScrollable: Boolean = node.config.contains(SemanticsActions.ScrollBy)
+
+    val isSelected: Boolean =
+        node.config.getOrNull(SemanticsProperties.Selected) == true
+
+    val className: String =
+        node.config.getOrNull(SemanticsProperties.Role)?.toString() ?: COMPOSE_TAG_NAME
+
+    val isPassword: Boolean =
+        node.config.getOrNull(SemanticsProperties.Password)?.let { true } ?: false
```
mykola-mokhnach, line 53: == true

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -88,10 +89,12 @@ class SourceDocument constructor(
 
     private fun setAttribute(attrName: ViewAttributesEnum, attrValue: Any?) {
         // Do not write attributes, whose values equal to null
-        attrValue?.let {
+        attrValue.let {
```
mykola-mokhnach, line 92: attrValue must not be nullable to apply such change

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -269,6 +335,11 @@ class SourceDocument constructor(
     }
 
     fun findViewsByXPath(xpathSelector: String): List<View> {
+        val indices = matchingNodeIds(xpathSelector, VIEW_INDEX)
```
mykola-mokhnach, line 338: this method could be made to oneliner

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/SourceDocument.kt
```diff
@@ -138,35 +141,35 @@ class SourceDocument constructor(
...
+            ViewAttributesEnum.ADAPTERS to null,
+            ViewAttributesEnum.ADAPTER_TYPE to null
         ).forEach {
             when (it.key) {
                 ViewAttributesEnum.TEXT, ViewAttributesEnum.HINT ->
                     if (!isTextOrHintRecorded && isAttributeIncluded(it.key)) {
-                        viewElement.text?.let { text ->
-                            setAttribute(ViewAttributesEnum.TEXT, text.rawText)
-                            setAttribute(ViewAttributesEnum.HINT, text.isHint)
+                        viewElement.text.let { text ->
```
mykola-mokhnach, line 170: is `text` non-nullable?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ComposeNodeFinder.kt
```diff
@@ -44,16 +44,27 @@ fun semanticsMatcherForLocator(locator: Locator): SemanticsMatcher =
...
+    }
+}
+
 fun getSemanticsNode(elementId: String): SemanticsNode =
     try {
         getNodeInteractionById(elementId).fetchSemanticsNode()
     } catch (e: AssertionError) {
         throw StaleElementException(elementId)
-    }
+    }
```
mykola-mokhnach, line 70: it is ok to have a line break at the end of a file

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ComposeNodeFinder.kt
```diff
@@ -44,16 +44,27 @@ fun semanticsMatcherForLocator(locator: Locator): SemanticsMatcher =
...
         Strategy.ACCESSIBILITY_ID -> hasContentDescription(locator.value!!)
+        Strategy.XPATH -> hasXpath(locator)
         else -> throw InvalidSelectorException(
             "Can't use non-Compose selectors. " +
                     "Only ${Strategy.VIEW_TAG}, ${Strategy.TEXT}, ${Strategy.LINK_TEXT} and " +
                     "${Strategy.ACCESSIBILITY_ID} are supported"
         )
     }
 
+private fun hasXpath(locator: Locator): SemanticsMatcher {
```
mykola-mokhnach, line 55: 👍

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/XMLHelpers.kt
```diff
@@ -21,36 +21,43 @@ import java.util.regex.Pattern
...
+            .replaceAll("")
         return XML10_TAG_PATTERN
-                .matcher(nodeName)
-                .replaceAll("_")
+            .matcher(nodeName)
+            .replaceAll("_")
     }
 
     fun toSafeString(source: Any?, replacement: String): String? {
         return if (source == null)
```
mykola-mokhnach, line 56: we could remove return and made it to an oneliner

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/XMLHelpers.kt
rajdeepv: It does change, Instead of no nodes, it adds nodes with empty values like `text="" `  in the XML.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: `node.config.getOrNull(SemanticsProperties.Password)` is not of Boolean type

mykola-mokhnach: then it could be `value?.toBoolean() ?: false`

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/XMLHelpers.kt
mykola-mokhnach: @KazuCocoa How do you think, would it be ok if we change that behaviour (no attribute vs empty-string attribute for null values) now or it would be better to keep it as it was for espresso?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
rajdeepv: `node.config.getOrNull(SemanticsProperties.Password)` returns `Unit` and doesn't respond to `toBoolean()`

mykola-mokhnach: oh, this is weird why it does not return boolean. Then the current implementation is fine

rajdeepv: I could actually replace it with `node.config.contains(SemanticsProperties.Password)`

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/XMLHelpers.kt
KazuCocoa: mm, empty-string can consider the element is set as empty-string, so it possibly confuses users if the attribute is set as empty or does not exist. btw, then, can XPath locator find the element with an empty string `text`?

I'd prefer to keep the previous case (null) to explicit no-attributes

mykola-mokhnach: btw in uiautomator2 and xcuitest we do transform nulls to empty strings

KazuCocoa: oh, i see...
So, then, are https://github.com/appium/appium-uiautomator2-driver/#element-attributes empty strings when they have no values? (I think we should address this point.)

Then an empty string is fine, but I think it would be nice to create another PR to leave the breaking change as a separate message. (appium 1.22 has been released, so we will not have such a central changelog for now.)

mykola-mokhnach: basically, how it works there, in the xml we replace null values with empty strings, but for `getAttribute` call nulls are still allowed

KazuCocoa: okay, it makes sense to me > but for getAttribute call nulls are still allowed

I see the replacement.
I'm ok to change this behavior to make the same as other drivers. But I still would like to make this change as a separate one from this compose feature.

rajdeepv: I think the same. In fact, we should reconsider changing the behavior of other drivers (no text is not the same as empty text)
For now, it is best to keep the compose behavior in line with espresso to avoid detour

test/functional/commands/jetpack-compose-e2e-specs.js
```diff
@@ -45,6 +45,19 @@ describe('Jetpack Compose', function () {
...
+    await driver.updateSettings({ driver: 'espresso' });
+    let el = await driver.elementByXPath("//*[@text='Clickable Component']");
+    await driver.moveTo(el);
+    await el.click();
+
+    await driver.updateSettings({ driver: 'compose' });
+
+    let e = await driver.elementByXPath("//*[@view-tag='lol']");
+    await e.text().should.eventually.equal('Click to see dialog');
+    await driver.back();
```
mykola-mokhnach, line 58: driver.back should be called in a `finally` block to prevent following tests from failing

mykola-mokhnach (approved):

KazuCocoa (approved):

test/functional/commands/jetpack-compose-e2e-specs.js
```diff
@@ -17,6 +17,11 @@ describe('Jetpack Compose', function () {
     }
     driver = await initSession(COMPOSE_CAPS);
   });
+
+  afterEach(async function () {
+    await driver.back();
```
mykola-mokhnach, line 22: ('m not quite sure this is going to work. `afterEach` is executed if any on the test steps fail. What if, for example, 
```
   let el = await driver.elementByXPath("//*[@text='Clickable Component']");
    await driver.moveTo(el);
    await el.click();
```
one of these steps fail? Then returning back would redirect us to a home screen.

If we use afterEach then beforeEach should also have appropriate content

rajdeepv: Yes, I borrowed it from `mobile-e2e-specs.js`. I could not find any example with `finally` in any of tests. Having interdependent tests is bad and probably we should switch to creating driver in before each and destroying it in after each.

test/functional/commands/jetpack-compose-e2e-specs.js
```diff
@@ -10,14 +10,15 @@ describe('Jetpack Compose', function () {
   this.timeout(MOCHA_TIMEOUT);
 
   let driver;
-  before(async function () {
+  beforeEach(async function () {
     // For SDK 23 and below Jetpack compose app crashes while running under instrumentation.
     if (parseInt(process.env.ANDROID_SDK_VERSION, 10) <= 23) {
```
mykola-mokhnach, line 15: skipping the suite might still be kept in `before` handler

test/functional/commands/jetpack-compose-e2e-specs.js
```diff
@@ -10,14 +10,15 @@ describe('Jetpack Compose', function () {
...
       return this.skip();
     }
     driver = await initSession(COMPOSE_CAPS);
```
mykola-mokhnach, line 18: this might add to the suite duration, but I'm fine with that, since we don't have many tests in this suite

PR #697, opened 2021-10-06
fix: avoid recursive serialization when result is non primitive

rajdeepv:
https://github.com/appium/appium/issues/15933


rajdeepv:
@mykola-mokhnach this might break the existing tests for people. I will fix our tests to have methods`[{name: "getTypeface"},{name: "getStyle"}]`

PR #700, opened 2021-10-11
feat: [JetPack Compose] Clear and Enter Keys Routes

sravanmedarapu:
Changes include:
- Clear, Enter Keys routes implementation and corresponding e2e tests
- In favour of less code and code changes backward tracking updated `RequestHandler` interface to following
             - RequestHandler interface will have abstract espresoo(`handleEspresso`) and compose (`handleCompose`)handler implementations.
             - Handler Implementation classes will implement both `handleEspresso`(existing handleInternal method will be rename to handleEspresso or similar) and `handleCompose`
              - `RequestHandler::invokeStrategy()` will decide which method/handler to invoke based current strategy type.

E2e Tests Depends on https://github.com/appium/appium-espresso-driver/pull/699

PR #702, opened 2021-10-13
fix: [compose] Fix class names in page source

sravanmedarapu:
Problem Statement:
- Currently implementation of page source returns element node as "ComposeNode" for most of the elements(except for [Role](https://github.com/androidx/androidx/blob/fc686a1579c3e1d5b1680906835effb4598566b5/compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/semantics/SemanticsProperties.kt#L544))
- Current [Logic](https://github.com/appium/appium-espresso-driver/blob/master/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt#L59)
e.g:
```
<ComposeNode bounds="[130,383][234,426]" class="ComposeNode" clickable="false" enabled="true" focused="false" index="0" password="false" properties="" resource-id="2623" scrollable="false" selected="false" text="Games"/>

<ComposeNode bounds="[150,322][213,385]" class="ComposeNode" clickable="false" content-desc="Favorite" enabled="true" focused="false" index="1" password="false" properties="" resource-id="2622" scrollable="false" selected="false"/>
```

Proposed Solution:
- As a better alternative, we can relay on semantic node properties to predict the element type and treat that as a class name.
- Using this approach above element nodes would look like
```
<Text bounds="[130,383][234,426]" class="Text" clickable="false" enabled="true" focused="false" index="0" password="false" properties="" resource-id="2623" scrollable="false" selected="false" text="Games"/>

<Image bounds="[150,322][213,385]" class="Image" clickable="false" content-desc="Favorite" enabled="true" focused="false" index="1" password="false" properties="" resource-id="2622" scrollable="false" selected="false"/>
```

Note:
Please review only after merging https://github.com/appium/appium-espresso-driver/pull/700, seems I created from wrong branch.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
@@ -17,12 +17,15 @@
 package io.appium.espressoserver.lib.model
 
 import android.graphics.Rect
-import androidx.compose.ui.semantics.*
+import androidx.compose.ui.semantics.SemanticsActions
+import androidx.compose.ui.semantics.SemanticsNode
+import androidx.compose.ui.semantics.SemanticsProperties
+import androidx.compose.ui.semantics.getOrNull
 import androidx.compose.ui.state.ToggleableState
 import io.appium.espressoserver.lib.handlers.exceptions.NotYetImplementedException
 import io.appium.espressoserver.lib.model.Rect.Companion.fromBounds
 
-const val COMPOSE_TAG_NAME = "ComposeNode"
+const val DEFAULT_TAG_NAME = "ComposeNode"
 
 class ComposeNodeElement(private val node: SemanticsNode) {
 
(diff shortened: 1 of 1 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -84,6 +84,20 @@ class ComposeNodeElement(private val node: SemanticsNode) {
     val rect: io.appium.espressoserver.lib.model.Rect
         get() = fromBounds(bounds)
 
+    fun getClassName(node: SemanticsNode): String {
```
mykola-mokhnach, line 87: shouldn't this method be private?

mykola-mokhnach: how about `guessClassName` ?

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -84,6 +84,20 @@ class ComposeNodeElement(private val node: SemanticsNode) {
...
+//        Following are the list of possible properties.
+        val propertiesList = listOf(SemanticsProperties.EditableText,
+            SemanticsProperties.Text,
+            SemanticsProperties.ProgressBarRangeInfo,
+            SemanticsProperties.HorizontalScrollAxisRange,
+            SemanticsProperties.VerticalScrollAxisRange,
+            SemanticsProperties.SelectableGroup,
+            SemanticsProperties.PaneTitle,
+            SemanticsProperties.Password)
+        return propertiesList.firstOrNull{  node.config.contains(it) }?.name ?: COMPOSE_TAG_NAME
```
mykola-mokhnach, line 98: extra space before `node`

KazuCocoa (approved):

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/ComposeNodeElement.kt
```diff
@@ -56,7 +59,22 @@ class ComposeNodeElement(private val node: SemanticsNode) {
         get() = node.config.getOrNull(SemanticsProperties.Selected) == true
 
     val className: String
-        get() = node.config.getOrNull(SemanticsProperties.Role)?.toString() ?: COMPOSE_TAG_NAME
+        get() {
+            //  Compose API doesn't have class name info, as a workaround relaying on node SemanticsProperties.
+            val possibleClassProperties = listOf(
```
mykola-mokhnach, line 64: can we move this list to a constant?

sravanmedarapu: Seems we can't non primitive types as constant, am I missing something?

`Const 'val' has type 'List<SemanticsPropertyKey<out Any>>'. Only primitives and String are allowed```

mykola-mokhnach: we could make a [lazy](https://kotlinlang.org/docs/delegated-properties.html#standard-delegates) private property for it, so the value is only initialized once.

mykola-mokhnach (approved):

Issue #713, opened 2021-11-05
Couldn't use .apks in appium espresso driver


skarindra:
Hi, I saw that in the Readme, we're allowed to use either `.apk` or `.apks` for the `app` capability, but when I tried to run the test using .apks, I got this error 

> org.openqa.selenium.SessionNotCreatedException: Unable to create a new remote session. Please check the server log for more details. Original error: A new session could not be created. Details: New app path '/Users/sakarindra/project/gitlab/marketplace-qa.apks' did not have extension: .apk

I'm using Appium 1.22.0.

Please advise, thank you!

mykola-mokhnach:
This doc is misleading. Espresso driver does not support files in .apks format. I've created a PR to actualize the README

skarindra:
noted, thanks @mykola-mokhnach ! Is there any plan for improvement, maybe in the future, to support `.apks` or `.aab` for espresso driver?

fadeltd:
> This doc is misleading. Espresso driver does not support files in .apks format. I've created a PR to actualize the README

is there any plan in the near future to improve appium espresso driver so it can support installed apks from [bundletool](https://github.com/google/bundletool)?

KazuCocoa:
let me test as https://github.com/appium/appium-espresso-driver/pull/714#issuecomment-962056964 this weekend with https://github.com/KazuCocoa/AppBundleSample.

But as the pr comments, we assume espresso needs more tricks to work with `.apks`.
As the updated readme, we think a single `.apk` generated by `bundletool` as `universal` for testing should work.

rajdeepv:
We can use .apks with espresso driver. After unzipping .apks, use splits/base-master.apk to get the package and launch activity. Also signing info can be provided by the user to create espresso-server and it will work.

KazuCocoa:
I've tested with a sample app with `.apks` (not universal one)
You can try https://github.com/appium/appium-espresso-driver/pull/718 on your local if this works for you.

skarindra:
@KazuCocoa I tried your solutions with below config : 

- Appium v1.22
- Appium Espresso Driver v1.48 (from your branch)
- As for the app, I generated the `.apks` with Appium keys

the app is successfully installed in my emulator but it seems like the espresso server couldn't be started, I keep getting error `socket hang up` and here's what I got from Appium log

[...]

PR #714, opened 2021-11-05
docs: Fix erroneous app description

mykola-mokhnach:
Espresso driver, actually, does not support .apks format. Related to https://github.com/appium/appium-espresso-driver/issues/713


KazuCocoa (approved):

KazuCocoa:
Probably can we support `.apks` once we add the ext in https://github.com/appium/appium-espresso-driver/blob/006bcfcf5a4d07713cb3e6d7a032ce80704cf28d/lib/driver.js#L177 ?

Let me test later if only the change help.

mykola-mokhnach:
@KazuCocoa I assume it won't work. The apk then must be patched with espresso server stuff to actually be able to run a test

KazuCocoa:
yea, like this, probably like this customization will be needed. (need to ensure this tho)
mm, we need a better testing app for the appbundle.
https://stackoverflow.com/questions/60709733/create-unit-and-instrumentation-test-for-android-project-with-dynamic-features

rajdeepv:
We can use `.apks` with espresso driver.  After unzipping `.apks`, use `splits/base-master.apk` to get the package and launch activity. Also signing info can be provided by the user to create espresso-server and it will work.

PR #715, opened 2021-11-05
chore: Bump Gradle version and address compilation warnings

mykola-mokhnach:


KazuCocoa:
oh, interesting...
```
* What went wrong:
An exception occurred applying plugin request [id: 'com.android.application']
> Failed to apply plugin 'com.android.internal.application'.
   > Android Gradle plugin requires Java 11 to run. You are currently using Java 1.8.
     You can try some of the following options:
       - changing the IDE settings.
       - changing the JAVA_HOME environment variable.
       - changing `org.gradle.java.home` in `gradle.properties`.
```

mykola-mokhnach:
@KazuCocoa It looks like Gradle 7 requires min Java11 to be enabled to successfully compile the sources. This might become an issue while using an older Android SDK (e.g. below 30). So the question is whether we should bump Gradle version now or to wait longer...

mykola-mokhnach:
It looks like we must use v11 if we want to continue with Jetpack testing, which makes sense...

mykola-mokhnach:
https://stackoverflow.com/questions/68728918/error-android-gradle-plugin-requires-java-11-to-run-you-are-currently-using-ja

KazuCocoa:
https://developer.android.com/studio/releases/gradle-plugin#jdk-11

> Incompatibility with 1.4.x Kotlin Multiplatform plugin
> Android Gradle Plugin 7.0.0 is compatible with Kotlin Multiplatform plugin 1.5.0 and higher. Projects that use the Kotlin Multiplatform support need to update to Kotlin 1.5.0 to use Android Gradle Plugin 7.0.0. As a workaround, you can downgrade the Android Gradle plugin to 4.2.x, although this is not recommended.

---

Yea... I read the above documentation after seeing this PR's CI error.

mm, `espressoBuildConfig` has `androidGradlePlugin`. So, for now, probably it would be nice to leave this v7 and Java 11 requirement in the gradle plugin section https://github.com/appium/appium-espresso-driver#espresso-build-config , and keep v4 as this project default...?

KazuCocoa:
> if we want to continue with Jetpack testing

ah...

KazuCocoa:
mm, then, it looks like we must bump this major version and address requirements about java 11 and build tools 31+ for the version. We also can address the lower appium-espresso version for gradle plugin v4 and lower build tools version.
(Then, we should create a version before bumping the major version)


For now, this newer espresso version is available only via `appium@beta` (not in 1.22.0 and .1) or appium 2.0.
Appium 2.0 can specify its driver version independently, so users who do not need compose can select the lower version for a while...

mykola-mokhnach:
I did split this PR, so we could apply features/fixes without bumping the major module version. Perhaps the deprecation of Appium1 would be the best time to publish it

KazuCocoa:
Good idea

README.md
```diff
@@ -20,9 +20,9 @@ The key difference between [UiAutomator2 Driver](https://github.com/appium/appiu
 On top of standard Appium requirements Espresso driver also expects the following prerequisites:
 
 - Windows, Linux and macOS are supported as hosts
-- [Android SDK Platform tools](https://developer.android.com/studio/releases/platform-tools) must be installed. [Android Studio IDE](https://developer.android.com/studio) also provides a convenient UI to install and manage the tools.
+- [Android SDK Platform tools](https://developer.android.com/studio/releases/platform-tools) must be installed. The minium required JDK version for the Espresso driver version 2 is `30.0.0`. For earlier driver versions older tools could still be used. [Android Studio IDE](https://developer.android.com/studio) also provides a convenient UI to install and manage the tools.
```
rajdeepv, line 23: The minium required `Android SDK` version

KazuCocoa:
note:
- GH actions' base JDK should be `11`
- The commit message should have `breaking change` message. Then, the auto release will bump the major version.

mykola-mokhnach:
@KazuCocoa I assume it would be ok to merge this change without bumping the major version, since we did it before to while making this driver compatible with Appium2

KazuCocoa:
ah, yea. It's true

PR #718, opened 2021-11-06
[do not merge yet] feat: allow apks

KazuCocoa:
I just tested as https://github.com/KazuCocoa/AppBundleSample#for-espresso-driver.
https://github.com/KazuCocoa/AppBundleSample/blob/master/test.rb passed.

So, for now, https://github.com/KazuCocoa/AppBundleSample#for-espresso-driver signs the modules in `apks` with appium's keys in appium-adb. (I've generated https://github.com/KazuCocoa/AppBundleSample/blob/master/apks/appium-ks from them)

Then, if the app under test has been a simple one, it worked.
We can ask https://github.com/appium/appium-espresso-driver/issues/713 in their cases.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -101,6 +101,7 @@ const CHROME_NO_PROXY = [
 ];
 
 const APP_EXTENSION = '.apk';
+const APPS_EXTENSION = '.apks';
 
 
 class EspressoDriver extends BaseDriver {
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
From the architectural perspective it would make sense to make configureApp call from base driver more flexible, so we could add some post-processing steps there for the given file types. I could do such change, but I'm not very motivated to backport it to Appium2 then, so I'd rather wait until v1 is deprecated and then only add it to the base driver v2.

KazuCocoa:
It is reasonable. Agree with it >  I'd rather wait until v1 is deprecated

KazuCocoa:
What about this?

Some apps like a test app such as https://github.com/KazuCocoa/AppBundleSample#for-espresso-driver should work with this well as same as UIA2, but some ones probably face the issue already mykola addressed.

PR #735, opened 2022-01-23
refactor: Switch the package to use Appium2 modules

mykola-mokhnach:


test/functional/commands/touch-e2e-specs.js
```diff
@@ -411,134 +411,134 @@ describe('touch actions -', function () {
 
   describe('mjsonwp touch actions', function () {
     describe('multi touch actions', function () {
```
jlipps, line 413: why not just add skips and a todo so we know what needs to happen here?

jlipps (approved):

KazuCocoa (approved):

test/functional/commands/touch-e2e-specs.js
mykola-mokhnach: these tests use wd imports. Leving them uncommented would require to rewrite the source to use wdio instead.

PR #750, opened 2022-02-15
feat: Add support of .aab packages

mykola-mokhnach:
The idea is simple. If an .aab package is provided we transform it to a universal .apk using bundletool and cache this .apk, because it is too expensive to do such action for each test if the same app is being tested (which is usually the case)


KazuCocoa (approved):
seems good, overall.
Should we append description in `appium:app` in README?
So, users have two options for `.aab`. One is by them, and the other is appium converts it to universal apk by this pr.

mykola-mokhnach:
Updated readme

KazuCocoa (approved):

lib/driver.js
```diff
@@ -219,6 +233,46 @@ class EspressoDriver extends BaseDriver {
     }
   }
 
+  async onPostConfigureApp ({cachedAppInfo, isUrl, appPath}) {
```
KazuCocoa, line 236: noticed still this `appPath` could have `.zip` extension, so `.zip` file always handles as `this.adb.extractUniversalApk` even it is `apk` after unzip.
https://github.com/appium/ruby_lib_core/issues/370

KazuCocoa: https://github.com/appium/appium/pull/16548

Issue #752, opened 2022-02-18
Windows: Escape ampersand in keystore password


headsvk:
Hi all.
Has anybody managed to put `&` into `keystorePassword` or `keyPassword` capability on Windows?

I'm getting build errors when building Espresso server with password like `START&END`:
```
Error: Unable to build Espresso server - Process ended with exitcode 1 (cmd: 'gradlew.bat -PappiumKeyPassword\=START&END')
Gradle error message:
END is not recognized as an internal or external command,
```

I tried to escape the password like `"START&END"` or `START\&END` or `START^&END` with no success.
It's unfortunate the password includes an ampersand but I cannot change it.

KazuCocoa:
What about giving the value as `'START"&"END'`?

headsvk:
@KazuCocoa that adds quotes around the whole parameter and more escaping. Runs as
`"-PappiumKeystorePassword='/START\"&\"END='"`  and fails with `'\"END'"'` is not recognized as an internal or external command.

KazuCocoa:
mm, what env did you run?
I checked some commands as https://github.com/appium/appium-espresso-driver/pull/753 (e.g. `./gradlew.bat -PappiumKeyPassword\='START\"&\"END'` did not show recognized error etc) on Win 10 powershell. (not via appium at this time)

Appium drivers are going to https://appiumpro.com/editions/122-installing-appium-20-and-the-driver-and-plugins-cli base, so you can install an arbitrary espresso driver as `appium driver install espresso@version`.

headsvk:
I understood to try it in Appium, sorry. Powershell needs more escaping for other params `./gradlew.bat -PappiumAndroidGradlePlugin\=7.1.1` fails with `Task '.1.1' not found in root project`. But that's unrelated, Appium uses cmd from what I see.

Running `gradlew.bat -PappiumKeyPassword\='START\"&\"END'` in cmd doesn't fail on recognized error but I get ` Keystore was tampered with, or password was incorrect`.


What works in cmd is simply `gradlew.bat -PappiumKeystorePassword="START&END"` but that cannot be done through Appium 1.22.0 because it wraps the whole argument in single quotes `gradlew.bat '-PappiumKeystorePassword="START&END"'`.

Thanks for the link, I will look into upgrading to 2.0.0 to keep up-to-date with drivers 🙂

KazuCocoa:
[...]

PR #753, opened 2022-02-19
fix: add double quote for gradle build config

KazuCocoa:
Fixes https://github.com/appium/appium-espresso-driver/issues/752

```
# before
% ./gradlew -PappiumKeyPassword=START&ENDStarting a Gradle Daemon (subsequent builds will be faster)

> Starting Daemon

%
[1]  + suspended (tty input)  ./gradlew -PappiumKeyPassword=START

# after
% ./gradlew -PappiumKeyPassword="START&END"
Starting a Gradle Daemon, 1 busy Daemon could not be reused, use --status for details
```

### DIFF
--- a/lib/server-builder.js
+++ b/lib/server-builder.js
@@ -74,7 +74,7 @@ class ServerBuilder {
 
   getCommand () {
     const cmd = system.isWindows() ? 'gradlew.bat' : './gradlew';
-    const buildProperty = (key, value) => value ? `-P${key}=${value}` : null;
+    const buildProperty = (key, value) => value ? `-P${key}="${value}"` : null;
     let args = VERSION_KEYS
       .filter((key) => key !== GRADLE_VERSION_KEY)
       .map((key) => {
--- a/test/unit/server-builder-specs.js
+++ b/test/unit/server-builder-specs.js
@@ -19,7 +19,7 @@ describe('server-builder', function () {
     });
 
     it('should pass only specified versions as properties and pass them correctly', function () {
-      const expected = {cmd: expectedCmd, args: ['-PappiumAndroidGradlePlugin=1.2.3', 'app:assembleAndroidTest']};
+      const expected = {cmd: expectedCmd, args: ['-PappiumAndroidGradlePlugin="1.2.3"', 'app:assembleAndroidTest']};
       let serverBuilder = new ServerBuilder({
         buildConfiguration: {
           toolsVersions: {


lib/server-builder.js
```diff
@@ -74,7 +74,7 @@ class ServerBuilder {
 
   getCommand () {
     const cmd = system.isWindows() ? 'gradlew.bat' : './gradlew';
-    const buildProperty = (key, value) => value ? `-P${key}=${value}` : null;
+    const buildProperty = (key, value) => value ? `-P${key}="${value}"` : null;
```
mykola-mokhnach, line 77: is this only needed on Windows?
what is going to happen if the value already contains quotes?

KazuCocoa: > is this only needed on Windows?

No, macOS also got the same error with `&` without `"` or `'`. The description's error was on macOS.

On Windows' case was (the error message was in Japanese tho) in powershell:
```
PS C:\Users\fly49\Downloads\appium-espresso-driver-master\appium-espresso-driver-master\espresso-server> ./gradlew.bat -PappiumKeyPassword=START&END
発生場所 行:1 文字:40
+ ./gradlew.bat -PappiumKeyPassword=START&END
+                                        ~
アンパサンド (&) 文字は許可されていません。& 演算子は、今後使用するために予約されています。アンパサンドを二重引用符で囲
み ("&")、文字列の一部として渡してください。
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : AmpersandNotAllowed
```

The above command worked in command prompt in Windows though.

`"xxxx\"yyyy"` or `"xxxx"yyyy"` like this string case the argument is divided with `"xxx"` and `"yyy"`. So in this case, we can use `'` intead. But probably no good way to avoid `"` and `'` at the same time.

KazuCocoa: mm, i need more dig Win env... changed to draft.

KazuCocoa: Hm, possibly we should set such value as ` ./gradlew.bat -PappiumKeyPassword\='START\"&\"END'` for Windows additionally. `"` itself needs in general.
Asking in #752

mykola-mokhnach: you could probably use a check similar to https://github.com/appium/appium-adb/blob/86807a3dbde31c4c2ae49e0fddfeafad368d6bdc/lib/tools/adb-commands.js#L579

KazuCocoa: It seems not all, but it seems safe to escape characters in ADB_INPUT_ESCAPE_PATTERN with `\"`. (at least, I got error messages by `<>|&`). So, chars in the pattern will be `'START\"&\"END'` (`\"`).

mykola-mokhnach, line 77: I would rather check explicitly if the corresponding value already contains quotes and throw an error.

KazuCocoa:
ok... so, for now, this PR can focus on wrapping arguments with `"` only. https://github.com/appium/appium-espresso-driver/issues/752#issuecomment-1047605525

lib/server-builder.js
KazuCocoa: It seems like the root cause was `-PappiumTargetPackage="io.appium.android.apis"` etc was wrapped by `'` in the command. So, with `"`, the command will be like:

```
[Espresso] Unable to build Espresso server - Process ended with exitcode 1 (cmd: './gradlew '-PappiumTargetPackage="io.appium.android.apis"' app\:assembleAndroidTest')
```

But expected one should be `./gradlew -PappiumTargetPackage="io.appium.android.apis" app\:assembleAndroidTest`

KazuCocoa: okay, i figured out.

The command will be `./gradlew '-PappiumKeyPassword=START&END'`.
Then, macOS and Linux have no issue. They handle "START&END" as a string.
But on Windows, `./gradlew '-PappiumKeyPassword=START&END'` will be `./gradlew -PappiumKeyPassword=START` and `END` in the command prompt. in the case, the argument should be `./gradlew -PappiumKeyPassword="START&END"`. In powershell, it should be `./gradlew -PappiumKeyPassword="START\&END"` (escape)

mykola-mokhnach: @KazuCocoa The gradle [documentation](https://docs.gradle.org/current/userguide/command_line_interface.html) says 

> Options that accept values can be specified with or without = between the option and argument; however, use of = is recommended.

So we could maybe try to make the value to a separate arg instead of quoting the whole expression

KazuCocoa: hm, I tried like the below command, but then, the separated arg was handled as a task.
```
% ./gradlew -PappiumKeyPassword "START&END" app:assembleAndroidTest 

FAILURE: Build failed with an exception.

* What went wrong:
Task 'START&END' not found in root project 'espresso-server'.

* Try:
Run gradlew tasks to get a list of available tasks. Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```


I've combined args as a command like:

```
    const gradlebuild = new SubProcess(cmd + ' ' + args.join(' '),[], {
      cwd: this.serverPath,
      stdio: ['ignore', 'pipe', 'pipe'],
    });
```
, then, the generated command was `./gradlew -PappiumTargetPackage="io.appium.android.apis" app:assembleAndroidTest ENOENT`. So the `ENOENT` prevent the command. Maybe i need to see `SubProcess`

lib/server-builder.js
```diff
@@ -163,6 +163,7 @@ class ServerBuilder {
     const gradlebuild = new SubProcess(cmd, args, {
       cwd: this.serverPath,
       stdio: ['ignore', 'pipe', 'pipe'],
+      windowsVerbatimArguments: true
```
KazuCocoa, line 166: May help...?
https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

KazuCocoa: https://github.com/appium/appium-espresso-driver/issues/752#issuecomment-1053029534 
Ok, so, with this one, at least `'` in Windows env will gone. Then, they can provide `"xxxx"` as the capability to avoid 
 the Windows's escape error....

KazuCocoa: So, with this PR and `"START&END"` as the capability, the 
```
END is not recognized as an internal or external command,
```
error should fix.

KazuCocoa: I'd like to add `"` automatically on Windows case after this merge and ask the reporter to try this

PR #758, opened 2022-03-11
fix: unzip if the file was zipped

KazuCocoa:
`.zip` ext local file has not been unzipped since v2.1.0.
To keep compatibility and `this.helpers.configureApp`'s behavior, it would be nice to unzip in `onPostConfigureApp` if the given local file was zipped one.
Then, `onPostConfigureApp` should return a new unzipped file as its response to give the unzipped file to set it as `this.opts.app` for the following process.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -261,8 +261,6 @@ class EspressoDriver extends BaseDriver {
       cwd: tmpRoot,
       strict: false,
     })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
-    logger.info(sortedBundleItems);
-
     const unzippedAppPath = path.join(tmpRoot, _.first(sortedBundleItems));
     logger.info(`'${unzippedAppPath}' is an unzipped file from '${appPath}'`);
     return unzippedAppPath;
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -260,9 +260,19 @@ class EspressoDriver extends BaseDriver {
       cwd: tmpRoot,
       strict: false,
     })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
-    const unzippedAppPath = path.resolve(tmpRoot, path.basename(_.first(sortedBundleItems)));
-    logger.info(`'${unzippedAppPath}' is the unzipped file from '${appPath}'`);
+    logger.info(sortedBundleItems);
+
+    const unzippedAppPath = path.join(tmpRoot, _.first(sortedBundleItems));
+    logger.info(`'${unzippedAppPath}' is an unzipped file from '${appPath}'`);
     return unzippedAppPath;
+  }
+
+  isApk (appPath) {
+    return _.endsWith(_.toLower(appPath), '.apk');
+  }
+
+  async extractUniversalApk (shouldExtract, appPath) {
+    return shouldExtract ? appPath : await this.adb.extractUniversalApk(appPath);
   }
 
   async onPostConfigureApp ({cachedAppInfo, isUrl, appPath}) {
@@ -276,28 +286,33 @@ class EspressoDriver extends BaseDriver {
       }
     };
 
-    const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
-    if (unzippedAppPath) {
-      appPath = unzippedAppPath;
-    }
-
-    const isApk = _.endsWith(_.toLower(appPath), '.apk');
-    // Do not cache .apk if was already present on the local file system
-    const shouldResultAppPathBeCached = !isApk || (isApk && isUrl);
     let pathInCache = null;
     let isResultAppPathAlreadyCached = false;
     if (_.isPlainObject(cachedAppInfo)) {
       const packageHash = await fs.hash(appPath);
       if (packageHash === cachedAppInfo.packageHash && await fs.exists(cachedAppInfo.fullPath)) {
+        logger.info(`Using '${cachedAppInfo.fullPath}' that is cached from '${appPath}'`);
         isResultAppPathAlreadyCached = true;
         pathInCache = cachedAppInfo.fullPath;
       }
     }
+
+    // appPath can be .zip, .apk, .apks and .aab
+    const isApk = this.isApk(appPath);
+    // Do not cache .apk if was already present on the local file system
+    const shouldResultAppPathBeCached = !isApk || (isApk && isUrl);
+
     if (!isResultAppPathAlreadyCached) {
       if (shouldResultAppPathBeCached) {
-        pathInCache = isApk ? appPath : await this.adb.extractUniversalApk(appPath);
+        // .zip, .aab or downloaded .apk
+        const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
+        const isZipAppPath = unzippedAppPath != null;
+        pathInCache = isZipAppPath
+          ? await this.extractUniversalApk(this.isApk(unzippedAppPath), unzippedAppPath)
+          : await this.extractUniversalApk(isApk, appPath);
+
         if (!isApk && isUrl) {
-          // Clean up the temporarily downloaded .aab package
+          // Clean up the temporarily downloaded one
           await fs.rimraf(appPath);
         }
         await presignApp(pathInCache);
@@ -308,11 +323,7 @@ class EspressoDriver extends BaseDriver {
       }
     }
 
-    if (shouldResultAppPathBeCached) {
-      return {appPath: pathInCache};
-    }
-
-    return unzippedAppPath ? {appPath} : false;
+    return shouldResultAppPathBeCached ? {appPath: pathInCache} : false;
   }
 
   get driverData () {
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -237,6 +237,34 @@ class EspressoDriver extends BaseDriver {
     }
   }
 
+  /**
+   * Unzip the given app path if needed.
+   *
+   * @param {string} appPath The path to app file.
+   * @returns {?string} Retuns the path to an unzipped app file path
+   *                    if the given path was unzipped, otherwise null.
+   */
+  async unzipAppIfNeeded (appPath) {
+    if (path.extname(_.toLower(appPath)) !== ZIP_EXT) {
+      return null;
+    }
+
+    const useSystemUnzipEnv = process.env.APPIUM_PREFER_SYSTEM_UNZIP;
+    const useSystemUnzip = _.isEmpty(useSystemUnzipEnv)
+      || !['0', 'false'].includes(_.toLower(useSystemUnzipEnv));
+    const tmpRoot = await tempDir.openDir();
+    await zip.extractAllTo(appPath, tmpRoot, {useSystemUnzip});
+
+    const globPattern = `**/*.+(${SUPPORTED_EXTENSIONS.map((ext) => ext.replace(/^\./, '')).join('|')})`;
+    const sortedBundleItems = (await fs.glob(globPattern, {
+      cwd: tmpRoot,
+      strict: false,
+    })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
+    const unzippedAppPath = path.resolve(tmpRoot, path.basename(_.first(sortedBundleItems)));
+    logger.info(`'${unzippedAppPath}' is the unzipped file from '${appPath}'`);
+    return unzippedAppPath;
+  }
+
   async onPostConfigureApp ({cachedAppInfo, isUrl, appPath}) {
     const presignApp = async (appLocation) => {
       if (this.opts.noSign) {
@@ -248,24 +276,9 @@ class EspressoDriver extends BaseDriver {
       }
     };
 
-    let isUnzippedAppPath = false;
-    if (path.extname(_.toLower(appPath)) === ZIP_EXT) {
-      const useSystemUnzipEnv = process.env.APPIUM_PREFER_SYSTEM_UNZIP;
-      const useSystemUnzip = _.isEmpty(useSystemUnzipEnv)
-        || !['0', 'false'].includes(_.toLower(useSystemUnzipEnv));
-      const tmpRoot = await tempDir.openDir();
-      await zip.extractAllTo(appPath, tmpRoot, {useSystemUnzip});
-
-      const globPattern = `**/*.+(${SUPPORTED_EXTENSIONS.map((ext) => ext.replace(/^\./, '')).join('|')})`;
-      const sortedBundleItems = (await fs.glob(globPattern, {
-        cwd: tmpRoot,
-        strict: false,
-      })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
-      const unzippedAppPath = path.resolve(tmpRoot, path.basename(_.first(sortedBundleItems)));
-      logger.info(`'${unzippedAppPath}' is the unzipped file from '${appPath}'`);
-      // The appPath is used in the following steps
+    const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
+    if (unzippedAppPath) {
       appPath = unzippedAppPath;
-      isUnzippedAppPath = true;
     }
 
     const isApk = _.endsWith(_.toLower(appPath), '.apk');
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -244,6 +248,26 @@ class EspressoDriver extends BaseDriver {
       }
     };
 
+    let isUnzippedAppPath = false;
```
mykola-mokhnach, line 251: can we move this block of code to a separate helper function to make the current one easier to parse?

lib/driver.js
```diff
@@ -270,7 +294,9 @@ class EspressoDriver extends BaseDriver {
         await presignApp(appPath);
       }
     }
-    return shouldResultAppPathBeCached ? {appPath: pathInCache} : false;
+
+    return shouldResultAppPathBeCached ?
```
mykola-mokhnach, line 298: nested ternary conditions are hard to read. Can we add a simple if condition instead?

lib/driver.js
KazuCocoa: moved to a method -> 4178724c312ee669c02407d7f1b68d4ac6e12a6e

lib/driver.js
```diff
@@ -244,6 +276,11 @@ class EspressoDriver extends BaseDriver {
       }
     };
 
+    const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
```
mykola-mokhnach, line 279: Please merge the unzipping logging with the caching logic below. Basically we must not unzip an app again if it was already cached and the app must be properly presigned.

lib/driver.js
```diff
@@ -233,6 +237,34 @@ class EspressoDriver extends BaseDriver {
...
+      || !['0', 'false'].includes(_.toLower(useSystemUnzipEnv));
+    const tmpRoot = await tempDir.openDir();
+    await zip.extractAllTo(appPath, tmpRoot, {useSystemUnzip});
+
+    const globPattern = `**/*.+(${SUPPORTED_EXTENSIONS.map((ext) => ext.replace(/^\./, '')).join('|')})`;
+    const sortedBundleItems = (await fs.glob(globPattern, {
+      cwd: tmpRoot,
+      strict: false,
+    })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
+    const unzippedAppPath = path.resolve(tmpRoot, path.basename(_.first(sortedBundleItems)));
```
mykola-mokhnach, line 263: I'm not quite sure `path.basename` should be used here as the destination file might also be found in a subfolder. We should first verify if any item has been matched across `sortedBundleItems` and then join a relative path to the first one with tmpRoot as a result or throw an exception

lib/driver.js
```diff
@@ -233,6 +238,42 @@ class EspressoDriver extends BaseDriver {
...
+      || !['0', 'false'].includes(_.toLower(useSystemUnzipEnv));
+    const tmpRoot = await tempDir.openDir();
+    await zip.extractAllTo(appPath, tmpRoot, {useSystemUnzip});
+
+    const globPattern = `**/*.+(${SUPPORTED_EXTENSIONS.map((ext) => ext.replace(/^\./, '')).join('|')})`;
+    const sortedBundleItems = (await fs.glob(globPattern, {
+      cwd: tmpRoot,
+      strict: false,
+    })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
+    const unzippedAppPath = path.join(tmpRoot, _.first(sortedBundleItems));
```
mykola-mokhnach, line 264: there is still no verification for the case `sortedBundleItems` is empty

lib/driver.js
```diff
@@ -244,21 +285,31 @@ class EspressoDriver extends BaseDriver {
...
+    const isApk = this.isApk(appPath);
+    // Do not cache .apk if was already present on the local file system
+    const shouldResultAppPathBeCached = !isApk || (isApk && isUrl);
+
     if (!isResultAppPathAlreadyCached) {
       if (shouldResultAppPathBeCached) {
-        pathInCache = isApk ? appPath : await this.adb.extractUniversalApk(appPath);
+        // .zip, .aab or downloaded .apk
+        const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
+        const isZipAppPath = unzippedAppPath != null;
```
mykola-mokhnach, line 308: could also be `!!unzippedAppPath`

lib/driver.js
```diff
@@ -233,6 +238,42 @@ class EspressoDriver extends BaseDriver {
...
+      cwd: tmpRoot,
+      strict: false,
+    })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
+    const unzippedAppPath = path.join(tmpRoot, _.first(sortedBundleItems));
+    logger.info(`'${unzippedAppPath}' is an unzipped file from '${appPath}'`);
+    return unzippedAppPath;
+  }
+
+  isApk (appPath) {
+    return _.endsWith(_.toLower(appPath), APK_EXT);
```
mykola-mokhnach, line 270: this should not necessarily be a class method

lib/driver.js
```diff
@@ -244,21 +285,31 @@ class EspressoDriver extends BaseDriver {
...
-        pathInCache = isApk ? appPath : await this.adb.extractUniversalApk(appPath);
+        // .zip, .aab or downloaded .apk
+        const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
+        const isZipAppPath = unzippedAppPath != null;
+        pathInCache = isZipAppPath
+          ? await this.extractUniversalApk(this.isApk(unzippedAppPath), unzippedAppPath)
+          : await this.extractUniversalApk(isApk, appPath);
+
         if (!isApk && isUrl) {
           // Clean up the temporarily downloaded .aab package
```
mykola-mokhnach, line 314: I assume the comment must be updated (.zip will also be deleted if downloaded)

lib/driver.js
```diff
@@ -244,21 +285,31 @@ class EspressoDriver extends BaseDriver {
...
+        logger.info(`Using '${cachedAppInfo.fullPath}' that is cached from '${appPath}'`);
         isResultAppPathAlreadyCached = true;
         pathInCache = cachedAppInfo.fullPath;
       }
     }
+
+    // appPath can be .zip, .apk, .apks and .aab
+    const isApk = this.isApk(appPath);
+    // Do not cache .apk if was already present on the local file system
+    const shouldResultAppPathBeCached = !isApk || (isApk && isUrl);
```
mykola-mokhnach, line 302: I assume shouldResultAppPathBeCached to be true for resulting .apk extracted from a zip archive

lib/driver.js
KazuCocoa: added when the zip does not have valid extension packages case

lib/driver.js
```diff
@@ -244,23 +281,47 @@ class EspressoDriver extends BaseDriver {
...
       if (shouldResultAppPathBeCached) {
-        pathInCache = isApk ? appPath : await this.adb.extractUniversalApk(appPath);
+        // .zip, .aab or downloaded .apk
+        const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
+        const isZipAppPath = !!unzippedAppPath;
+
+        // Probably the appPath zip file did not have SUPPORTED_EXTENSIONS.
+        if (hasZipkExt(appPath) && !isZipAppPath) {
+          logger.errorAndThrow(`${this.opts.app} did not have any of '${SUPPORTED_EXTENSIONS.join(', ')}' ` +
+            `extension packages. Please make sure the zip file has a valid package.`);
```
KazuCocoa, line 315: e.g.
> Selenium::WebDriver::Error::SessionNotCreatedError: A new session could not be created. Details: /Users/kazuaki/GitHub/ruby_lib_core/test/functional/app/a.zip did not have any of '.apk, .aab' extension packages. Please make sure the zip file has a valid package. Check https://github.com/appium/appium-espresso-driver#troubleshooting regarding advanced session startup troubleshooting.

lib/driver.js
```diff
@@ -244,23 +281,47 @@ class EspressoDriver extends BaseDriver {
...
+    const isApk = hasApkExt(appPath);
+    // local .zip, .aab and downloaded .zip, .aab and .apks should be cached
+    // to skip unzip and generate .apk from .aab.
+    const shouldResultAppPathBeCached = !isApk || (isApk && isUrl);
+
     if (!isResultAppPathAlreadyCached) {
       if (shouldResultAppPathBeCached) {
-        pathInCache = isApk ? appPath : await this.adb.extractUniversalApk(appPath);
+        // .zip, .aab or downloaded .apk
+        const unzippedAppPath = await this.unzipAppIfNeeded(appPath);
```
mykola-mokhnach, line 309: what if we make this more explicit:

```
if (ath.extname(_.toLower(appPath)) === ZIP_EXT) {
  appPath = await unzipApp()...
}
```

also, if no apps were found in the archive then an exception could be thrown directly un the unzipApp method.

Then we could rid of the complex validation below

lib/driver.js
```diff
@@ -244,23 +281,47 @@ class EspressoDriver extends BaseDriver {
...
+          logger.errorAndThrow(`${this.opts.app} did not have any of '${SUPPORTED_EXTENSIONS.join(', ')}' ` +
+            `extension packages. Please make sure the zip file has a valid package.`);
+        }
+
+        // unzippedAppPath or appPath has SUPPORTED_EXTENSIONS.
+        pathInCache = isUnzippedAppPath
+          ? await extractUniversalApk(hasApkExt(unzippedAppPath), unzippedAppPath)
+          : await extractUniversalApk(isApk, appPath);
+
         if (!isApk && isUrl) {
```
mykola-mokhnach, line 323: I assume the clenup won't work as expected if .zip archive contains .aab file. According to the current logic the .aab  will be extracted, then transformed to .apk, but it won't be cleaned up. Only the parent .zip is going to be deleted if it was not a local file in such case

lib/driver.js
KazuCocoa: sounds good.

lib/driver.js
KazuCocoa: Yes, you're correct. b9b348a cleans the local .aap that is unzipped one

lib/driver.js
```diff
@@ -233,6 +238,36 @@ class EspressoDriver extends BaseDriver {
...
+
+    const globPattern = `**/*.+(${SUPPORTED_EXTENSIONS.map((ext) => ext.replace(/^\./, '')).join('|')})`;
+    const sortedBundleItems = (await fs.glob(globPattern, {
+      cwd: tmpRoot,
+      strict: false,
+    })).sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
+    if (sortedBundleItems.length === 0) {
+      // no expected packages in the zip
+      logger.errorAndThrow(`${this.opts.app} did not have any of '${SUPPORTED_EXTENSIONS.join(', ')}' ` +
+      `extension packages. Please make sure the zip file has a valid package.`);
```
mykola-mokhnach, line 264: Please make sure the provided .zip archive contains at least one valid package bundle

lib/driver.js
```diff
@@ -233,6 +238,36 @@ class EspressoDriver extends BaseDriver {
...
+    }
+    const unzippedAppPath = path.join(tmpRoot, _.first(sortedBundleItems));
+    logger.info(`'${unzippedAppPath}' is an unzipped file from '${appPath}'`);
```
mykola-mokhnach, line 267: is the unzipped file

mykola-mokhnach: I'd say this is a debug-level log

lib/driver.js
```diff
@@ -244,25 +279,52 @@ class EspressoDriver extends BaseDriver {
...
+    const hasApkExt = (appPath) => _.endsWith(_.toLower(appPath), APK_EXT);
+    const extractUniversalApk = async (shouldExtract, appPath) =>
+      shouldExtract ? appPath : await this.adb.extractUniversalApk(appPath);
+
     let pathInCache = null;
     let isResultAppPathAlreadyCached = false;
     if (_.isPlainObject(cachedAppInfo)) {
       const packageHash = await fs.hash(appPath);
       if (packageHash === cachedAppInfo.packageHash && await fs.exists(cachedAppInfo.fullPath)) {
+        logger.info(`Using '${cachedAppInfo.fullPath}' that is cached from '${appPath}'`);
```
mykola-mokhnach, line 291: which is cached

lib/driver.js
```diff
@@ -244,25 +279,52 @@ class EspressoDriver extends BaseDriver {
...
         isResultAppPathAlreadyCached = true;
         pathInCache = cachedAppInfo.fullPath;
       }
     }
+
+    // appPath can be .zip, .apk, .apks or .aab
```
mykola-mokhnach, line 297: .apks is not supported

lib/driver.js
```diff
@@ -244,25 +279,52 @@ class EspressoDriver extends BaseDriver {
...
+        pathInCache = isUnzippedAppPath
+          ? await extractUniversalApk(isUnzippedApk, unzippedAppPath)
+          : await extractUniversalApk(isApk, appPath);
+
         if (!isApk && isUrl) {
-          // Clean up the temporarily downloaded .aab package
+          // Clean up the temporarily downloaded .aab or .zip package
           await fs.rimraf(appPath);
         }
+        if (!isUnzippedApk) {
```
mykola-mokhnach, line 324: I'd rather explicitly check whether unzippedAppPath has an .aab extension

lib/driver.js
```diff
@@ -244,25 +279,52 @@ class EspressoDriver extends BaseDriver {
...
+    // local .zip, .aab and downloaded .zip, .aab and .apks should be cached
+    // to skip unzip and generate .apk from .aab.
+    const shouldResultAppPathBeCached = !isApk || (isApk && isUrl);
+
     if (!isResultAppPathAlreadyCached) {
       if (shouldResultAppPathBeCached) {
-        pathInCache = isApk ? appPath : await this.adb.extractUniversalApk(appPath);
+        // .zip, .aab or downloaded .apk
+        let unzippedAppPath;
+        let isUnzippedAppPath = false;
```
mykola-mokhnach, line 307: do we need this intermediate variable? Cannot we use `unzippedAppPath` for this purpose?

KazuCocoa: not necessary. that was just my preference :)

KazuCocoa:
updated

lib/driver.js
```diff
@@ -244,25 +280,51 @@ class EspressoDriver extends BaseDriver {
...
+        // .zip, .aab or downloaded .apk
+        let unzippedAppPath;
+        let isUnzippedApk = false;
+        if (path.extname(_.toLower(appPath)) === ZIP_EXT) {
+          unzippedAppPath = await this.unzipApp(appPath);
+          isUnzippedApk = hasApkExt(unzippedAppPath);
+        }
+
+        // unzippedAppPath or appPath has SUPPORTED_EXTENSIONS.
+        pathInCache = _.isUndefined(unzippedAppPath)
```
mykola-mokhnach, line 316: I'd just use `unzippedAppPath` as a flag, without  `isUndefined ` check

lib/driver.js
```diff
@@ -244,25 +280,51 @@ class EspressoDriver extends BaseDriver {
...
+    // to skip unzip and generate .apk from .aab.
+    const shouldResultAppPathBeCached = !isApk || (isApk && isUrl);
+
     if (!isResultAppPathAlreadyCached) {
       if (shouldResultAppPathBeCached) {
-        pathInCache = isApk ? appPath : await this.adb.extractUniversalApk(appPath);
+        // .zip, .aab or downloaded .apk
+        let unzippedAppPath;
+        let isUnzippedApk = false;
+        if (path.extname(_.toLower(appPath)) === ZIP_EXT) {
```
mykola-mokhnach, line 310: this extension check could be unified with ones above (apk and aab)

lib/driver.js
```diff
@@ -244,25 +280,51 @@ class EspressoDriver extends BaseDriver {
...
       if (packageHash === cachedAppInfo.packageHash && await fs.exists(cachedAppInfo.fullPath)) {
+        logger.info(`Using '${cachedAppInfo.fullPath}' which is cached from '${appPath}'`);
         isResultAppPathAlreadyCached = true;
         pathInCache = cachedAppInfo.fullPath;
       }
     }
+
+    // appPath can be .zip, .apk or .aab
+    const isApk = hasApkExt(appPath);
+    // local .zip, .aab and downloaded .zip, .aab and .apk should be cached
```
mykola-mokhnach, line 301: This comment is hard to parse. Could be simplified to:

```Only local .apk files that are available in-place should not be cached```

mykola-mokhnach (approved):

Issue #763, opened 2022-04-01
Error: Unable to build Espresso server - spawn ./gradlew ENOENT


zpiao1:
Hi when I try to run the espresso server I got the ENOENT error. 

# Logs
```
[Espresso] Unable to build Espresso server - spawn ./gradlew ENOENT
[Espresso] Gradle error message:
[Espresso] 
[debug] [Espresso] Deleting espresso session
[debug] [W3C] Encountered internal error running command: A new session could not be created. Details: Unable to build Espresso server - spawn ./gradlew ENOENT
[debug] [W3C] Gradle error message:
[debug] [W3C] . Check https://github.com/appium/appium-espresso-driver#troubleshooting regarding advanced session startup troubleshooting.
[debug] [W3C] Error: Unable to build Espresso server - spawn ./gradlew ENOENT
[debug] [W3C] Gradle error message:
[debug] [W3C] . Check https://github.com/appium/appium-espresso-driver#troubleshooting regarding advanced session startup troubleshooting.
[debug] [W3C]     at Object.wrappedLogger.errorAndThrow (/Users/zhaojy/ui-test/node_modules/@appium/support/lib/logging.js:94:35)
[debug] [W3C]     at ServerBuilder.runBuildProcess (/Users/zhaojy/ui-test/node_modules/appium-espresso-driver/lib/server-builder.js:195:4)
[debug] [W3C]     at ServerBuilder.build (/Users/zhaojy/ui-test/node_modules/appium-espresso-driver/lib/server-builder.js:73:5)
[debug] [W3C]     at EspressoRunner.buildNewModServer (/Users/zhaojy/ui-test/node_modules/appium-espresso-driver/lib/espresso-runner.js:188:5)
[debug] [W3C]     at EspressoRunner.installTestApk (/Users/zhaojy/ui-test/node_modules/appium-espresso-driver/lib/espresso-runner.js:146:7)
[debug] [W3C]     at EspressoDriver.initAUT (/Users/zhaojy/ui-test/node_modules/appium-espresso-driver/lib/driver.js:552:7)
[debug] [W3C]     at EspressoDriver.startEspressoSession (/Users/zhaojy/ui-test/node_modules/appium-espresso-driver/lib/driver.js:422:5)
[debug] [W3C]     at EspressoDriver.createSession (/Users/zhaojy/ui-test/node_modules/appium-espresso-driver/lib/driver.js:225:7)
[...]

PR #764, opened 2022-04-01
fix: set gradle command as full path

KazuCocoa:
https://github.com/appium/appium-espresso-driver/issues/763

### DIFF
--- a/lib/server-builder.js
+++ b/lib/server-builder.js
@@ -72,7 +72,7 @@ class ServerBuilder {
   }
 
   getCommand () {
-    const cmd = system.isWindows() ? 'gradlew.bat' : './gradlew';
+    const cmd = path.resolve(this.serverPath, system.isWindows() ? 'gradlew.bat' : 'gradlew');
     const buildProperty = (key, value) => value ? `-P${key}=${value}` : null;
     let args = VERSION_KEYS
       .filter((key) => key !== GRADLE_VERSION_KEY)
--- a/lib/server-builder.js
+++ b/lib/server-builder.js
@@ -72,7 +72,7 @@ class ServerBuilder {
   }
 
   getCommand () {
-    const cmd = path.resolve(this.serverPath, system.isWindows() ? 'gradlew.bat' : 'gradlew');
+    const cmd = system.isWindows() ? 'gradlew.bat' : path.resolve(this.serverPath, 'gradlew');
     const buildProperty = (key, value) => value ? `-P${key}=${value}` : null;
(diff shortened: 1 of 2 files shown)


mykola-mokhnach (approved):

lib/server-builder.js
```diff
@@ -72,7 +72,7 @@ class ServerBuilder {
   }
 
   getCommand () {
-    const cmd = system.isWindows() ? 'gradlew.bat' : './gradlew';
+    const cmd = path.resolve(this.serverPath, system.isWindows() ? 'gradlew.bat' : 'gradlew');
```
mykola-mokhnach, line 75: I'm not quite sure this is going to work as expected, at least on Windows where .bat files are not executable on their own.
I assume it would be safer to run these scripts using `cmd.exe /c` on windows and using `bash` on other systems

KazuCocoa: oh, i see. > win

Yes, bash might be safe in case users change the permission by some reasons...
Let me keep .bat for Win in this PR.

I'll create another pr to add bash and cmd.exe to separate versions.

KazuCocoa: The below full path worked in both powershell and command prompt on Win 10, but `cmd.exe /c` did not in the command prompt. (Access denied error occurred). Powershell worked with the `/c`.

So, on Win, it seems like only with `gradlew.bat` or the full path to `gradlew.bat` is safer.

```
C:\Users\kazu\Downloads\appium-espresso-driver-master\appium-espresso-driver-master\espresso-server>C:\kazu\kazu\Downloads\appium-espresso-driver-master\appium-espresso-driver-master\espresso-server\gradlew.bat

FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\kazu\Downloads\appium-espresso-driver-master\appium-espresso-driver-master\espresso-server\app\build.gradle.kts' line: 1

* What went wrong:
An exception occurred applying plugin request [id: 'com.android.application']
> Failed to apply plugin 'com.android.internal.application'.
   > Android Gradle plugin requires Java 11 to run. You are currently using Java 1.8.
     You can try some of the following options:
       - changing the IDE settings.
       - changing the JAVA_HOME environment variable.
       - changing `org.gradle.java.home` in `gradle.properties`.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
<-------------> 0% WAITING
> IDLE
```

PR #782, opened 2022-05-17
chore: Support `checked` attribute

sravanmedarapu:
**Issue:**
```
io.appium.espressoserver.lib.handlers.exceptions.NotYetImplementedException: Compose doesn't support attribute 'checked', Attribute name should be one of 'bounds', 'class', 'clickable', 'content-desc', 'enabled', 'focused', 'index', 'password', 'resource-id', 'scrollable', 'selected', 'text', 'view-tag'
        at io.appium.espressoserver.lib.model.ComposeNodeElement.getAttribute(ComposeNodeElement.kt:126)
```

**Reason:** 
**Checked** attribute not been respected for Compose Driver.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/AttributesEnum.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/AttributesEnum.kt
@@ -59,7 +59,6 @@ class EspressoAttributes : CommonAttributes(commonAttributes + espressoAttribute
                 AttributesEnum.ADAPTERS,
                 AttributesEnum.ADAPTER_TYPE,
                 AttributesEnum.CHECKABLE,
-                AttributesEnum.CHECKED,
                 AttributesEnum.FOCUSABLE,
                 AttributesEnum.HINT,
                 AttributesEnum.INSTANCE,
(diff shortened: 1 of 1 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/model/AttributesEnum.kt
```diff
@@ -106,6 +106,6 @@ abstract class CommonAttributes(var attributes: List<AttributesEnum>) {
     }
 
     fun valueOf(value: String): AttributesEnum? {
-        return attributes.find { it.toString().equals(value, ignoreCase = false) }
+        return attributes.find { it.toString().equals(value, ignoreCase = true) }
```
mykola-mokhnach, line 109: what is the reason to do this change?

sravanmedarapu: Checking the case didn't make sense to me. I can revert back if needed, but not sure why we need to respect case in this case.
e.g.: "CHECKED" == "checked"

mykola-mokhnach: because attribute names in xml are in lowercase. Also the change itself has nothing to with this PR

sravanmedarapu: Could see your point, reverted.

mykola-mokhnach (approved):

PR #784, opened 2022-05-29
feat: Add compose support more for element

KazuCocoa:
https://github.com/appium/appium/discussions/17001#discussioncomment-2842792

GetXXXX via elements like displayed has compose handler while some of them had only espresso ones.
This PR adds other GetXXXX for compose.

```
class GetDisplayed : RequestHandler<AppiumParams, Boolean> {

    override fun handleEspresso(params: AppiumParams): Boolean =
        ViewElement(EspressoElement.getViewById(params.elementId, false)).isVisible

    override fun handleCompose(params: AppiumParams): Boolean =
        try {
            getNodeInteractionById(params.elementId).assertIsDisplayed()
            true
        } catch (e: AssertionError) {
            false
        }
}
```

These implementation basically follows existing `class Text : RequestHandler<AppiumParams, String>`'s implementation.

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetLocationInView.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetLocationInView.kt
@@ -16,18 +16,15 @@
 
 package io.appium.espressoserver.lib.handlers
 
-import io.appium.espressoserver.lib.handlers.exceptions.AppiumException
 import io.appium.espressoserver.lib.model.AppiumParams
-import io.appium.espressoserver.lib.model.EspressoElement
 import io.appium.espressoserver.lib.model.Location
+import io.appium.espressoserver.lib.model.EspressoElement
 import io.appium.espressoserver.lib.model.ViewElement
 
 class GetLocationInView : RequestHandler<AppiumParams, Location> {
 
-    @Throws(AppiumException::class)
-    override fun handleInternal(params: AppiumParams): Location {
-        val view = EspressoElement.getViewById(params.elementId)
-        val viewElement = ViewElement(view)
+    override fun handleEspresso(params: AppiumParams): Location {
+        val viewElement = ViewElement(EspressoElement.getViewById(params.elementId))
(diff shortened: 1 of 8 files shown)


rajdeepv (approved):

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetLocationInView.kt
```diff
@@ -16,18 +16,15 @@
...
 import io.appium.espressoserver.lib.model.AppiumParams
-import io.appium.espressoserver.lib.model.EspressoElement
 import io.appium.espressoserver.lib.model.Location
+import io.appium.espressoserver.lib.model.EspressoElement
 import io.appium.espressoserver.lib.model.ViewElement
 
 class GetLocationInView : RequestHandler<AppiumParams, Location> {
 
-    @Throws(AppiumException::class)
-    override fun handleInternal(params: AppiumParams): Location {
```
mykola-mokhnach, line 28: do we also want to have handleCompose here?

mykola-mokhnach:
should we also update the corresponding README section about newly available compose APIs?

attrideepak:
@KazuCocoa I have verified these changes..works fine now.

espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/handlers/GetLocationInView.kt
KazuCocoa: Perhaps yes, but I don't see the same method in the compose element so this PR does not have it so far.

mykola-mokhnach:
@KazuCocoa Would it also make sense to add some more e2e for the newly added Compose methods?

KazuCocoa:
Yes, am thinking to add these in https://github.com/appium/appium-espresso-driver/blob/master/test/functional/commands/jetpack-compose-attributes-e2e-specs.js as another pr.
I've added missing isSelected in the README.

mykola-mokhnach (approved):

KazuCocoa:
@attrideepak thank you for your help!

Issue #802, opened 2022-07-20
StartActivity does not support IntentOptions


msaravana123:
We are migrating UiAutomator2 scripts to Espresso.

I am facing issue with startingActivity with intentOptions (as Intent Action and Intent Arguments are not supported)

UiAutomator2:
``` 
Activity activity = new Activity(appPackage, appActivity);
activity.setIntentAction(intentAction);
activity.setOptionalIntentArguments(intentArguments);
driver.startActivity(activity);
```
Espresso
```
Activity activity = new Activity(appPackage, appActivity);
activity.setOptionalIntentArguments(intentArguments); //IntentOptions as JSON String
driver.startActivity(activity);
```

But only appPackage and appActivity are sent to Espresso server but intent options are not sent even though espresso startactivity supports IntentOptions.

The following call works and starts the correct activity with right data setting.
http://localhost:8300/session/{sessionId}/appium/device/start_activity

```
{
    "appPackage": "XXXXXX",
    "appActivity": "XXXXXXXXX",
    "optionalIntentArguments": {
        "flags": "ACTIVITY_NEW_TASK",
        "action": "XXXXXXXX",
        "className": "XXXXX",
        "es": {
            "username": "pilot1",
            "env": "Test",
            "featureFlags": "XXXXXXXXX",
            "env1": "XXXXX",
            "env2": "XXXXXXX"
        }
    }
}
```

msaravana123:
Raised a PR - https://github.com/appium/appium-espresso-driver/pull/803. Please review.

KazuCocoa:
https://github.com/appium/appium-espresso-driver/pull/803#issuecomment-1197730349

PR #803, opened 2022-07-20
feat: Add `mobile: startActivity` extension

msaravana123:
Update startActivity to support IntentOptions and passed as optionalIntentArguments. Clients might require update but this should help teams with both UiAutomator2 and Espresso scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -732,7 +732,7 @@ remotePath | string | yes | The full path to the remote file or a file inside an
 
 ### mobile: startActivity
 
-Starts the given activity with intent options, activity options and locale.
+Starts the given activity with intent options, activity options and locale. Activity could only be executed in scope of the current app package.
 
 #### Arguments
 
--- a/README.md
+++ b/README.md
@@ -730,6 +730,19 @@ Name | Type | Required | Description | Example
 --- | --- | --- | --- | ---
 remotePath | string | yes | The full path to the remote file or a file inside an application bundle | `/sdcard/myfile.txt` or `@my.app.id/path/in/bundle`
 
+### mobile: startActivity
+
+Starts the given activity with intent options, activity options and locale.
+
+#### Arguments
+
+Name | Type | Required | Description | Example
+--- | --- | --- | --- | ---
+appActivity | string | yes | Applicaiton acitivity identifier to start | `com.myapp.myacitivty`
+locale | object | no | Sets the locale for the app under test. It only uses public APIs for its purpose. See https://github.com/libyal/libfwnt/wiki/Language-Code-identifiers to get the list of available language abbreviations. | `{"language": "zh", "country": "CN", "variant": "Hans"}`
+optionalIntentArguments | object | no | The mapping of custom options for the intent that is going to be passed to the main app activity. Check [Intent Options](#intent-options) for more details. | `{ 'flags': 'ACTIVITY_NEW_TASK', 'action': '<intent_action>', 'className': '<fullyQualifiedAppActivity>', 'es': {'foo': 'bar'} }`
+optionalActivityArguments | object | no | The mapping of custom options for the main app activity that is going to be started. Check [Activity Options](#activity-options) for more details. | `{ 'launchDisplayId': 1 }`
+
 ### mobile: startService
 
 Starts the given service intent.
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -297,9 +297,9 @@ helpers.suspendChromedriverProxy = function suspendChromedriverProxy () {
 };
 
 /**
- *  Starts the given activity with intent options, activity options and locale. 
+ *  Starts the given activity with intent options, activity options and locale.
  *  Activity could only be executed in scope of the current app package.
- *  
+ *
  *  appActivity is mandatory
  *
  *  locale, optionalIntentArguments, optionalActivityArguments are optional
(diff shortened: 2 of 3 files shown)


KazuCocoa:
Thank you, CI is not happy. Perhaps tests related to startActivity needs updates.
Btw, did this `optionalIntentArguments` work in your env?

msaravana123:
> Thank you, CI is not happy. Perhaps tests related to startActivity needs updates. Btw, did this `optionalIntentArguments` work in your env?

Yes. optionalIntentArguments works in by env as espresso-server accepts the following payload and it works.

```
{
    "appPackage": "XXXXXX",
    "appActivity": "XXXXXXXXX",
    "optionalIntentArguments": {
        "flags": "ACTIVITY_NEW_TASK",
        "action": "XXXXXXXX",
        "className": "XXXXX",
        "es": {
            "username": "pilot1",
            "env": "Test",
            "featureFlags": "XXXXXXXXX",
            "env1": "XXXXX",
            "env2": "XXXXXXX"
        }
    }
}
```

lib/commands/general.js
```diff
@@ -297,17 +297,24 @@ helpers.suspendChromedriverProxy = function suspendChromedriverProxy () {
...
+  // intentAction, intentCategory, intentFlags, dontStopAppOnReset
   // parameters are not supported by Espresso
   appPackage = appPackage || this.caps.appPackage;
   appWaitPackage = appWaitPackage || appPackage;
   appActivity = qualifyActivityName(appActivity, appPackage);
   appWaitActivity = qualifyActivityName(appWaitActivity || appActivity, appWaitPackage);
+  try {
+    optionalIntentArguments = (optionalIntentArguments) ? JSON.parse(optionalIntentArguments) : null;
+  } catch (e) {
+    this.log.error('Cannot parse the optionalIntentArguments JSON', e);
```
mykola-mokhnach, line 310: what is the purpose in rethrowing of the exception?

lib/commands/general.js
```diff
@@ -297,17 +297,24 @@ helpers.suspendChromedriverProxy = function suspendChromedriverProxy () {
 };
 
 commands.startActivity = async function startActivity (appPackage, appActivity,
```
mykola-mokhnach, line 299: is there a documentation for this call that must be updated? AFAIK in clients with strict typing the optionalIntentArguments parameter has type "string" and it is going to be confusing to pass a json there. We either must do the appropriate transformation here or (what I would do) use the corresponding driver-specific `mobile:` extension instead

KazuCocoa: > optionalIntentArguments parameter has type "string" 

side note: 
https://github.com/appium/appium-android-driver/blob/4ec9994570bdf163b47b7318002a2d0b3687f128/lib/desired-caps.js#L132-L134 expects string

lib/commands/general.js
msaravana123: As mentioned in the issue https://github.com/appium/appium-espresso-driver/issues/802, I am passing intentoptions as JSON string and rethrowing the error to halt the execution if there are JSON parsing error and let user know that valid JSON is expected.

lib/commands/general.js
msaravana123: > corresponding driver-specific `mobile:` extension instead

I wanted to use mobile: startActivity, but is removed in espresso driver and hence passing the intent options as JSON string from client (as client expects and sends String).

But espresso server expects Intent Arguments as Map and hence doing the transformation here from string to JSON.

I can add a new mobile command instead if that is recommended like below

```
commands.mobileStartActivity = async function mobileStartActivity(opts = {}) {
  const {
    appPackage,
    appActivity,
    locale,
    optionalIntentArguments,
    optionalActivityArguments
  } = assertRequiredOptions(opts, ['appPackage', 'appActivity']);
  return await this.espresso.jwproxy.command(`/appium/device/start_activity`, 'POST', {
    appPackage,
    appActivity,
    locale,
    optionalIntentArguments,
    optionalActivityArguments
  });
```

Add `startActivity: 'mobileStartActivity',` to execute.js and will update the doc for expresso mobile:  comments section.

mykola-mokhnach: having a separate mobile command makes more sense to me, as we could provide a proper documentation for its arguments in driver README. Also, unlike in UIA2, with Espresso we cannot really customize `appPackage` value - it must be equal to the current app under test as Espresso driver only works inside of an instrumented app sandbox.

KazuCocoa (approved):

lib/commands/general.js
```diff
@@ -296,6 +296,23 @@ helpers.suspendChromedriverProxy = function suspendChromedriverProxy () {
   this.jwpProxyActive = true;
 };
 
+commands.mobileStartActivity = async function mobileStartActivity (opts = {}) {
```
mykola-mokhnach, line 299: Please also document opts here

README.md
```diff
@@ -730,6 +730,19 @@ Name | Type | Required | Description | Example
 --- | --- | --- | --- | ---
 remotePath | string | yes | The full path to the remote file or a file inside an application bundle | `/sdcard/myfile.txt` or `@my.app.id/path/in/bundle`
 
+### mobile: startActivity
+
+Starts the given activity with intent options, activity options and locale.
```
mykola-mokhnach, line 735: it makes sense to explicitly mention that the activity could only be executed in scope of the current app package

mykola-mokhnach (approved):
Had some minor comments. Otherwise LGTM

lib/commands/general.js
msaravana123: Added document

README.md
msaravana123: Added note about appPackage.

mykola-mokhnach (approved):

Issue #821, opened 2022-10-05
Is it possible to add a way to change the root of the DataMatcher?


Scrubles:
I'm trying to select a item on the dropdown of an AutoCompleteTextView. First I tried with the UiAutomator2, but there was nothing available there to make it work, then I tried with this library, but I stopped at the same point. I made it work running Espresso directly on my app, but I needed to use the method inRoot(isPlatformPopup()), which isn't available here, is it possible to add a way to call it?

I tried to call the inAdapterView, but I think it's 2 different things, and it didn't work.

KazuCocoa:
Maybe https://developer.android.com/reference/androidx/test/espresso/matcher/RootMatchers ?
Then I think currently this driver does not have it yet.

Scrubles:
Do you if there's any other way to find elements on popups?

mykola-mokhnach:
Have you tried to use [-android viewmatcher](https://github.com/appium/appium-espresso-driver#espresso-elements-location) locators?

KazuCocoa:
> First I tried with the UiAutomator2, but there was nothing available there to make it work

so, potentially this does not help, but you could use the below in an espresso session:
https://github.com/appium/appium-espresso-driver#mobile-uiautomator
https://github.com/appium/appium-espresso-driver#mobile-uiautomatorpagesource

mykola-mokhnach:
After v2.11.0 is is possible to provide root matchers to `-android viewmatcher` locators. See https://github.com/appium/appium-espresso-driver/pull/822

Scrubles:
I don't know if I should reopen this issue or create a new one.
Using the root matcher, I was able to find the element, but when I try to click on it, I receive this Exception.

[debug] [EspressoDriver@8d20 (f832bd3b)] Matched '/session/f832bd3b-3a58-4794-ae71-9c1d0bab2867/element' to command name 'findElement'
[...]

PR #824, opened 2022-10-13
fix: Properly fetch view interaction if it was found under a root

mykola-mokhnach:
Addresses https://github.com/appium/appium-espresso-driver/issues/821#issuecomment-1277909305

### DIFF
--- a/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.kt
+++ b/espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.kt
@@ -43,7 +43,6 @@ import androidx.test.espresso.Root
 import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.matcher.ViewMatchers.isDescendantOfA
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
-import androidx.test.espresso.matcher.ViewMatchers.withClassName
 import androidx.test.espresso.matcher.ViewMatchers.withContentDescription
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withTagValue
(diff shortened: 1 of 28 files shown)


espresso-server/app/src/androidTest/java/io/appium/espressoserver/lib/helpers/ViewFinder.kt
```diff
@@ -125,72 +125,72 @@ object ViewFinder {
...
                     // Perform Exception means nothing was found. Return empty list
                     emptyList()
                 }
             }
             Strategy.VIEWMATCHER -> {
                 val matcherJson = selector.toJsonMatcher()
-                views = try {
+                return try {
                     @Suppress("UNCHECKED_CAST")
                     getViews(root, matcherJson.query.matcher as Matcher<View>, findOne, matcherJson.query.scope as Matcher<Root>)
```
Scrubles, line 185: Shouldn't this scope be nullable?

mykola-mokhnach: it is nullable. Does something looks suspicious here?

Scrubles: If I not mistaken the cast should be Matcher<Root>?, when the scope is null it will throw an Exception

mykola-mokhnach: It should work just fine, `as?` is needed to do cast from types other than null: https://stackoverflow.com/questions/50836017/safe-cast-vs-cast-to-nullable

KazuCocoa (approved):

KazuCocoa (approved):
Maybe we can revert the test code change right now (until figured out this change was expected or not)

mykola-mokhnach:
ok, I'll revert the change. This is anyway an unrelated issue

PR #846, opened 2023-01-11
fix: specify supported non-standard commands in newMethodMap

jlipps:
depends on https://github.com/appium/appium-android-driver/pull/782

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -6,11 +6,12 @@ import { EspressoRunner, TEST_APK_PKG } from './espresso-runner';
 import { fs, tempDir, zip } from 'appium/support';
 import commands from './commands';
 import { DEFAULT_ADB_PORT } from 'appium-adb';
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -125,6 +126,17 @@ const AAB_EXT = '.aab';
 const SUPPORTED_EXTENSIONS = [APK_EXT, AAB_EXT];
 
 class EspressoDriver extends BaseDriver {
+
+  static newMethodMap = {
```
mykola-mokhnach, line 130: I'd also put it into separate file as in android driver. More entries could be added there later

jlipps: ok, done

.gitignore
```diff
@@ -14,3 +14,4 @@ node_modules
 build
 .DS_Store
 .gradle/
+add-method-map.js
```
mykola-mokhnach, line 17: not needed?

mykola-mokhnach (approved):

PR #854, opened 2023-02-04
chore: add print espresso path as a command

KazuCocoa:
similar to opening wda fir, I'd like to show where the espresso server is. This could be on Win as well, so this command shows where it is. This will help when a user wants to modify the gradle file directly in their debug for example. I had questions about where the espresso server was in past, so this command may help in such cases as well.

Example (local installed espresso driver dir)
```
kazu$ appium driver run --json espresso print-espresso-path
{
    "output": [
        "[STDOUT] /Users/kazu/GitHub/appium-espresso-driver/espresso-server"
    ]
}
kazu$ appium driver run espresso print-espresso-path
[STDOUT] /Users/kazu/GitHub/appium-espresso-driver/espresso-server
✔ log-espresso-path successfully ran
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ On top of standard Appium requirements Espresso driver also expects the followin
 
 ## Scripts
 
-- `appium driver run log-espresso-path` prints the path to the Appium Espresso server root. You can modify the gradle file directly if [Espresso Build Config](#espresso-build-config) was not sufficient.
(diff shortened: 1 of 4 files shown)


mykola-mokhnach (approved):

mykola-mokhnach:
@KazuCocoa Please do not forget to add the new `scripts` folder to https://github.com/appium/appium-espresso-driver/blob/aa0991cc504ec8456f40495a904898452535d73b/package.json#L50

Otherwise it won't be present in the published package

KazuCocoa:
yea, true. I forgot it..

Issue #534, opened 2020-01-16
Disabling animations by default results in unexpected consequence


rajdeepv:
As part of #73 Animations are completely disabled for espresso. This results in some unexpected effects like some images are not loaded (In my case even app logo was missing)
We can still have it disabled by default but there should be an option not to do so. I ran my test by reverting #73 and it worked fine

mykola-mokhnach:
I also like the idea to make the setting configurable

mykola-mokhnach:
and enabled by default

rajdeepv:
I tried with my app keeping animation on, did not face any issues. Maybe there can be other apps with never-ending animations that may not work.

KazuCocoa:
Thanks. It's interesting.

In my memory, a long time ago worked with animation, but espresso started to raise an exception (I forgot which version). But the limitation has been relaxed again.
Ah... or just an error message (not such exception..)

rajdeepv:
@mykola-mokhnach @KazuCocoa I ran the full suite of my app with around 400 tests with animation **ON** and everything looks good
I suggest keeping animation ON by default and utilize(implement) `disableWindowAnimation` capability to disable them if needed

nicodn:
Something else I've noticed that may relate to this is that the animation setting is permanently changed on the test device, it is **not** restored to whatever setting was configured prior to running the tests. I suggest that the driver restore the previous setting after testing is done, perhaps also controlled by a configurable capability.

rajdeepv:
@nicodn There is code that restores it. However, if the test is force killed in the middle, it permanently changes this setting.

pr4bh4sh:
This issue can be closed now.

PR #857, opened 2023-02-28
feat: add disableWindowAnimation caps

pr4bh4sh:
Address https://github.com/appium/appium-espresso-driver/issues/534 and makes animation configurable before launching the AUT.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
-appium:disableWindowAnimation | Disables the window animation for Android device under test. `true` by default
+appium:disableWindowAnimation | To avoid flakiness google [recommends](https://developer.android.com/training/testing/espresso/setup#set-up-environment) to disable the window animation of the Android device under test when running espresso test. `true` by default
 
 
 ## Settings API
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
-appium:setAnimation | Sets the animation for Android device under test. `false` by default
+appium:disableWindowAnimation | Disables the window animation for Android device under test. `true` by default
 
 
 ## Settings API
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
+appium:setAnimation | Sets the animation for Android device under test. `false` by default
 
 
 ## Settings API
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -394,10 +394,9 @@ class EspressoDriver extends BaseDriver {
     // TODO with multiple devices we'll need to parameterize this
     await helpers.initDevice(this.adb, this.opts);
 
-    // default state is window animation disabled
-    this.caps.disableWindowAnimation = this.caps.disableWindowAnimation !== false;
     // https://github.com/appium/appium-espresso-driver/issues/72
-    await this.setWindowAnimationState(this.caps.disableWindowAnimation);
+    // default state is window animation disabled
+    await this.setWindowAnimationState(this.caps.disableWindowAnimation !== false);
 
     // set actual device name, udid
     this.caps.deviceName = this.adb.curDeviceId;
@@ -459,10 +458,10 @@ class EspressoDriver extends BaseDriver {
     await this.addDeviceInfoToCaps();
   }
 
-  async setWindowAnimationState(disableAnimation) {
+  async setWindowAnimationState(animationState) {
     const isAnimationOn = await this.adb.isAnimationOn();
-    const shouldDisableAnimation = disableAnimation && isAnimationOn;
-    const shouldEnableAnimation = !disableAnimation && !isAnimationOn;
+    const shouldDisableAnimation = animationState && isAnimationOn;
+    const shouldEnableAnimation = !animationState && !isAnimationOn;
 
     if (shouldDisableAnimation) {
       this.log.debug('Disabling window animation as "disableWindowAnimation" capability is set to true');
@@ -473,7 +472,7 @@ class EspressoDriver extends BaseDriver {
       await this.adb.setAnimationState(true);
       this.wasAnimationEnabled = false;
     } else {
-      this.log.debug(`Window animation is already ${disableAnimation ? 'disabled' : 'enabled'}`);
+      this.log.debug(`Window animation is already ${animationState ? 'disabled' : 'enabled'}`);
     }
   }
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -179,6 +179,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
+appium:setAnimation | Sets the animation for Android device under test. `false` by default
```
KazuCocoa, line 182: Could we use `disableWindowAnimation` by default `true` to keep it consistent with UIA2 as well? (if this animation had no issue)

lib/driver.js
```diff
@@ -393,8 +393,19 @@ class EspressoDriver extends BaseDriver {
     // start an avd, set the language/locale, pick an emulator, etc...
     // TODO with multiple devices we'll need to parameterize this
     await helpers.initDevice(this.adb, this.opts);
+    this.caps.setAnimation = this.caps.setAnimation || false;
     // https://github.com/appium/appium-espresso-driver/issues/72
-    if (await this.adb.isAnimationOn()) {
+    if (this.caps.setAnimation && await this.adb.isAnimationOn()) {
```
mykola-mokhnach, line 398: consider only calling `await this.adb.isAnimationOn()`. Each extra adb call takes time.

lib/driver.js
```diff
@@ -393,8 +393,19 @@ class EspressoDriver extends BaseDriver {
...
+      this.log.debug('Animation is already on');
+      this.wasAnimationEnabled = true;
+    } else if (this.caps.setAnimation && (await this.adb.isAnimationOn() === false)) {
+      try {
+        await this.adb.setAnimationState(true);
+        this.wasAnimationEnabled = false;
+      } catch (err) {
+        this.log.warn(`Unable to turn on animations: ${err.message}`);
+      }
+    } else if (this.caps.setAnimation === false && await this.adb.isAnimationOn()) {
```
mykola-mokhnach, line 408: I don't see a need for an explicit `false` equality check after `this.caps.setAnimation = this.caps.setAnimation || false` has been done above. 

Also, is it really necessary to set it explicitly to false?

lib/driver.js
```diff
@@ -393,8 +393,19 @@ class EspressoDriver extends BaseDriver {
...
-    if (await this.adb.isAnimationOn()) {
+    if (this.caps.setAnimation && await this.adb.isAnimationOn()) {
+      this.log.debug('Animation is already on');
+      this.wasAnimationEnabled = true;
+    } else if (this.caps.setAnimation && (await this.adb.isAnimationOn() === false)) {
+      try {
+        await this.adb.setAnimationState(true);
+        this.wasAnimationEnabled = false;
+      } catch (err) {
+        this.log.warn(`Unable to turn on animations: ${err.message}`);
```
mykola-mokhnach, line 406: we should not silence this error if the animation enablement has been explicitly asked in capabilities.

lib/driver.js
```diff
@@ -393,8 +393,19 @@ class EspressoDriver extends BaseDriver {
     // start an avd, set the language/locale, pick an emulator, etc...
     // TODO with multiple devices we'll need to parameterize this
     await helpers.initDevice(this.adb, this.opts);
+    this.caps.setAnimation = this.caps.setAnimation || false;
     // https://github.com/appium/appium-espresso-driver/issues/72
-    if (await this.adb.isAnimationOn()) {
+    if (this.caps.setAnimation && await this.adb.isAnimationOn()) {
+      this.log.debug('Animation is already on');
+      this.wasAnimationEnabled = true;
```
mykola-mokhnach, line 400: if animation was already on then we should not change its state afterwards

lib/driver.js
```diff
@@ -393,8 +393,19 @@ class EspressoDriver extends BaseDriver {
     // start an avd, set the language/locale, pick an emulator, etc...
     // TODO with multiple devices we'll need to parameterize this
     await helpers.initDevice(this.adb, this.opts);
+    this.caps.setAnimation = this.caps.setAnimation || false;
```
mykola-mokhnach, line 396: maybe it makes sence to extract this logic into a separate method as it's getting a but more lengthy and complicated

lib/driver.js
pr4bh4sh: refactor, the state is true unless `disableWindowAnimation` is explicitly set to false.

lib/driver.js
pr4bh4sh: removed the try catch

lib/driver.js
pr4bh4sh: extracted into a function

lib/driver.js
```diff
@@ -393,15 +393,11 @@ class EspressoDriver extends BaseDriver {
     // start an avd, set the language/locale, pick an emulator, etc...
     // TODO with multiple devices we'll need to parameterize this
     await helpers.initDevice(this.adb, this.opts);
+
+    // default state is window animation disabled
+    this.caps.disableWindowAnimation = this.caps.disableWindowAnimation !== false;
```
mykola-mokhnach, line 398: I don't see a need to override the cap value. It is enough to call `await this.setWindowAnimationState(this.caps.disableWindowAnimation !== false);`

lib/driver.js
```diff
@@ -463,6 +459,24 @@ class EspressoDriver extends BaseDriver {
     await this.addDeviceInfoToCaps();
   }
 
+  async setWindowAnimationState(disableAnimation) {
```
mykola-mokhnach, line 462: the argument name seems confusing to me. I would rather change it to `isEnabled`

pr4bh4sh: the default state for the variable is `true`.
`disableAnimation === true` animation is disabled
`disableAnimation === false` animation is enabled

`isEnabled === true` sounds like animation should be enabled but when the cap `disableWindowAnimation` value is true animation should be disabled
and wise versa for the `false` value of `disableWindowAnimation`.

Should I just change it to `animationState`. We are anyway setting the value true by `this.caps.disableWindowAnimation !== false` except for the explicit `false` value.

mykola-mokhnach: it does not matter here the cap name is. It is about the arg name for this particular method. Eventually, `state` assumes multiple possible states, but this one could be on or off (true/false)

mykola-mokhnach: > isEnabled === true sounds like animation should be enabled but when the cap disableWindowAnimation value is true animation should be disabled

Right, simply pass `disableWindowAnimation === false` to it (inverted condition)

pr4bh4sh: Done

lib/driver.js
```diff
@@ -463,6 +458,24 @@ class EspressoDriver extends BaseDriver {
     await this.addDeviceInfoToCaps();
   }
 
+  async setWindowAnimationState(isEnabled) {
+    const isAnimationOn = await this.adb.isAnimationOn();
+    const shouldDisableAnimation = !isEnabled && isAnimationOn;
+    const shouldEnableAnimation = isEnabled && !isAnimationOn;
+
+    if (shouldDisableAnimation) {
+      this.log.debug('Disabling window animation as "disableWindowAnimation" capability is set to true');
```
mykola-mokhnach, line 467: it might not be necessarily set to true (e.g. if the default value is used)

mykola-mokhnach (approved):

README.md
```diff
@@ -179,6 +179,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
+appium:disableWindowAnimation | Disables the window animation for Android device under test. `true` by default
```
mykola-mokhnach, line 182: Perhaps it makes sense to describe in more details how enabling or disabling animations affects Espresso tests

KazuCocoa (approved):

lib/driver.js
pr4bh4sh: updated the message

README.md
pr4bh4sh: Added link for google espresso testing recommendation https://developer.android.com/training/testing/espresso/setup#set-up-environment

Issue #851, opened 2023-01-24
[Feature request] Can Appium espresso server apk be build without starting the Appium server?


pr4bh4sh:
I'm trying to find a stand-alone command to build the appium espresso server apk.

Currently, to build a server apk I use the below step.
1. Clone this repo and run `./gradlew clean assembleDebug assembleAndroidTest` inside `appium-espresso-driver/espresso-server` directory.
2. Run `npm install && npm link` in `appium-espresso-driver` directory.
3. Add the below additional caps in appium driver caps
   
```json
  noSign:               false,
  useKeystore:          true,
  keystorePath:         File.expand_path(<keystore_path>),
  keystorePassword:     KEYSTORE_PASSWORD
  keyAlias:             KEYSTORE_ALIAS,
  keyPassword:          KEYSTORE_PASSWORD,
  forceEspressoRebuild: true,
  showGradleLog:        true,
  espressoBuildConfig:    '{"toolsVersions": {"gradle": "7.5.1", "androidGradlePlugin": "7.3.1", "composeVersion": "1.3.1", "compileSdk": "33", "kotlin": "1.7.10"}, "additionalAndroidTestDependencies": ["androidx.compose.ui:ui-test-junit4:1.3.0"]}',
   ```
4. Execute a test so the appium espresso server apk would be generated.

I saw this Gradle command in the appium server log while creating the server apk

```bash
/var/folders/jj/0gsw19wd55n8vfrpmc6r961w0000gq/T/appium_temp/0/espresso-server-emulator-5556/gradlew 
-PappiumCompileSdk=32 
-PappiumKotlin=1.6.21 
-PappiumComposeVersion=1.2.1 
-PappiumZipAlign=true 
-PappiumKeystoreFile=pathh_to_kyastore
-PappiumKeystorePassword=password
-PappiumKeyAlias=keyalias 
-PappiumKeyPassword=password
-PappiumTargetPackage=<app_package>
app:assembleAndroidTest' 
in directory '/var/folders/jj/0gsw19wd55n8vfrpmc6r961w0000gq/T/appium_temp/0/espresso-server-emulator-5556'
```

Is there a command option where the capability can be passed as a command line argument?

or possibly a node js command line tool to build the appium espresso server which accepts the <u> keystore details, espressoBuildConfig </u> and builds the apk file.

cc @KazuCocoa

KazuCocoa:
Sounds good. Let me try

KazuCocoa:
[...]

PR #858, opened 2023-03-05
[wip] add espresso build command

KazuCocoa:
This is still wip. I'd like to extract the building espresso server via npm command over appium command, eventually.
https://github.com/appium/appium-espresso-driver#espresso-build-config

This does not change the gradle file itself, but it allows users to build the server with some parameters https://github.com/appium/appium-espresso-driver/issues/851

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "build": "npm run build:node && npm run build:server",
     "build:node": "rimraf build && babel --out-dir=build/lib lib && babel --out-dir=build index.js",
     "build:server": "cd espresso-server && ./gradlew clean assembleAndroidTest || cd ..",
+    "build:server_js": "node ./scripts/build-espressoserver.js",
     "dev": "npm run build -- --watch",
     "lint": "eslint .",
     "lint:server": "cd espresso-server && ./gradlew lint || cd ..",
--- a/scripts/build-espressoserver.js
+++ b/scripts/build-espressoserver.js
@@ -1,0 +1,29 @@
+const path = require('path');
+const { asyncify } = require('asyncbox');
+const { logger } = require('@appium/support');
+const { ServerBuilder } = require('../build/lib/server-builder.js');
+
+const LOG = new logger.getLogger('EspressoBuild');
+
+const ROOT_DIR = path.resolve(__dirname, '..');
+const ESPRESSO_SERVER_ROOT = path.join(ROOT_DIR, 'espresso-server');
+
+async function buildEspressoServer () {
+  const opts = {
+    serverPath: ESPRESSO_SERVER_ROOT,
+    showGradleLog: true,
+  };
+
+  if (process.env.TEST_APP_PACKAGE) {
+    opts.testAppPackage = process.env.TEST_APP_PACKAGE;
+  }
+
+  const builder = new ServerBuilder(LOG, opts);
+  await builder.build();
+}
+
+if (require.main === module) {
+  asyncify(buildEspressoServer);
+}
+
+module.exports = buildEspressoServer;


scripts/build-espresso.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  if (process.env.ESPRESSO_BUILD_CONFIG) {
+    if (!(await fs.exists(process.env.ESPRESSO_BUILD_CONFIG))) {
+      throw Error(`'${process.env.ESPRESSO_BUILD_CONFIG}' did not exist. Please set the path as an absolute path.`);
+    }
+    try {
+      const buildConfigurationStr = await fs.readFile(process.env.ESPRESSO_BUILD_CONFIG, 'utf8');
+      opts.buildConfiguration = JSON.parse(buildConfigurationStr);
+      console.log(`The espresso build config is ${JSON.stringify(opts.buildConfiguration)}`); // eslint-disable-line no-console
```
mykola-mokhnach, line 31: why cannot we just use the logger above to print it?

scripts/build-espresso.js
```diff
@@ -0,0 +1,44 @@
...
+const ESPRESSO_SERVER_ROOT = path.join(ROOT_DIR, 'espresso-server');
+
+const ESPRESSO_SERVER_BUILD = path.join(ESPRESSO_SERVER_ROOT, 'app', 'build');
+
+async function buildEspressoServer () {
+  console.log(`Deleting the build directory ${ESPRESSO_SERVER_BUILD}`); // eslint-disable-line no-console
+
+  const opts = {
+    serverPath: ESPRESSO_SERVER_ROOT,
+    showGradleLog: process.env.SHOW_GRADLE_LOG ? process.env.SHOW_GRADLE_LOG : false
```
mykola-mokhnach, line 17: not sure if we need to check the presence of env variables or also explicitly check if their values are not false/0

scripts/build-espresso.js
mykola-mokhnach: same above

scripts/build-espresso.js
```diff
@@ -0,0 +1,44 @@
...
+    } catch (e) {
+      throw Error(`Failed to parse the ${process.env.ESPRESSO_BUILD_CONFIG}. Please make sure that the JSON is valid format. Error: ${e}`);
+    }
+  }
+
+  const builder = new ServerBuilder(LOG, opts);
+  await builder.build();
+
+  const dstPath = path.resolve(ESPRESSO_SERVER_ROOT, 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug');
+  console.log(`The built server apk, app-debug-androidTest.apk, is in ${dstPath}`); // eslint-disable-line no-console
```
mykola-mokhnach, line 41: I would rather just show the full path together with the file name

scripts/build-espresso.js
```diff
@@ -0,0 +1,44 @@
...
+      const buildConfigurationStr = await fs.readFile(process.env.ESPRESSO_BUILD_CONFIG, 'utf8');
+      opts.buildConfiguration = JSON.parse(buildConfigurationStr);
+      console.log(`The espresso build config is ${JSON.stringify(opts.buildConfiguration)}`); // eslint-disable-line no-console
+    } catch (e) {
+      throw Error(`Failed to parse the ${process.env.ESPRESSO_BUILD_CONFIG}. Please make sure that the JSON is valid format. Error: ${e}`);
+    }
+  }
+
+  const builder = new ServerBuilder(LOG, opts);
+  await builder.build();
```
mykola-mokhnach, line 38: do we mirror the builder output?

mykola-mokhnach, line 38: we might provide some hint on enabling SHOW_GRADLE_LOG if the build fails

scripts/build-espresso.js
KazuCocoa: I considered the way as well, but I expected the primary usage was to open the path with Finder (on macOS for example) to refer to the apk instead of opening the apk itself.

Some terminal tools can open the path directly. e.g. cmd + click the link. When they open the path, if the path is `.apk`, the system will open an application instead of the Finder. Some tools require selecting the entire path and copy&paste it. This does not have much differences in the full path vs the file's directory.

Thus, at this time, showing the directory helps the expected use case with some tools, but other tools do not have differences to reach the .apk. So I did the current method.

I like the current way for this use case I expected, but I can put the full path instead.

scripts/build-espresso.js
KazuCocoa: 👍 [56b2d76](https://github.com/appium/appium-espresso-driver/pull/858/commits/56b2d768920a4e995b2d660a6df80c955b8a3327)

scripts/build-espresso.js
KazuCocoa: Changed to enable it only when the given env var was 1 or true. Other cases will be false as same as the default behavior.

scripts/build-espresso.js
```diff
@@ -0,0 +1,50 @@
+const l = require('lodash');
```
mykola-mokhnach, line 1: usually the variable is called just named as `_`

scripts/build-espresso.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const builder = new ServerBuilder(LOG, opts);
+  try {
+    await builder.build();
+  } catch (e) {
+    throw Error(`Failed to build the espresso server. SHOW_GRADLE_LOG=true environment variable helps to check the gradle log. Error: ${e}`);
+  }
+
+  const dstPath = path.resolve(ESPRESSO_SERVER_ROOT, 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug');
+  LOG.info(`The built server apk, app-debug-androidTest.apk, is in ${dstPath}`);
```
mykola-mokhnach, line 47: I meant the dstPath values should already contain `app-debug-androidTest.apk`

KazuCocoa: Ok, I've changed. Still I like to open the directory to open the link via cmd+click in a terminal, but no strong opinion for this.

scripts/build-espresso.js
```diff
@@ -0,0 +1,54 @@
...
+  if (process.env.ESPRESSO_BUILD_CONFIG) {
+    if (!(await fs.exists(process.env.ESPRESSO_BUILD_CONFIG))) {
+      throw Error(`'${process.env.ESPRESSO_BUILD_CONFIG}' did not exist. Please set the path as an absolute path.`);
+    }
+    try {
+      const buildConfigurationStr = await fs.readFile(process.env.ESPRESSO_BUILD_CONFIG, 'utf8');
+      opts.buildConfiguration = JSON.parse(buildConfigurationStr);
+      LOG.info(`The espresso build config is ${JSON.stringify(opts.buildConfiguration)}`);
+    } catch (e) {
+      throw Error(`Failed to parse the ${process.env.ESPRESSO_BUILD_CONFIG}. Please make sure that the JSON is valid format. Error: ${e}`);
```
mykola-mokhnach, line 35: these lines are too long. Please split them. Same below

scripts/build-espresso.js
```diff
@@ -0,0 +1,54 @@
...
+  const builder = new ServerBuilder(LOG, opts);
+  try {
+    await builder.build();
+  } catch (e) {
+    throw Error(`Failed to build the espresso server. SHOW_GRADLE_LOG=true environment variable helps to check the gradle log. Error: ${e}`);
+  }
+
+  const dstPath = path.resolve(ESPRESSO_SERVER_ROOT, 'app', 'build', 'outputs', 'apk', 'androidTest', 'debug', 'app-debug-androidTest.apk');
+  if (await fs.exists(dstPath)) {
+    LOG.info(`The built server apk is ${dstPath}`);
```
mykola-mokhnach, line 48: I'd rather say: `Full path to the server APK: ...`

scripts/build-espresso.js
```diff
@@ -0,0 +1,54 @@
...
+  } else {
+    LOG.info(`The built server androidTest.apk might be in ${ESPRESSO_SERVER_BUILD}`);
```
mykola-mokhnach, line 50: `Full path to the server build folder: ...`

README.md
```diff
@@ -42,7 +42,12 @@ On top of standard Appium requirements Espresso driver also expects the followin
 
 ## Scripts
 
-- `appium driver run print-espresso-path` prints the path to the Appium Espresso server root. You can modify the gradle file directly if [Espresso Build Config](#espresso-build-config) was not sufficient.
+- `appium driver run espresso print-espresso-path` prints the path to the Appium Espresso server root. You can modify the gradle file directly if [Espresso Build Config](#espresso-build-config) was not sufficient.
+- `appium driver run espresso build-espresso` builds the espresso server since driver version 2.18.0. It helps building the espresso server outside Appium process. Available environment variables are below:
```
mykola-mokhnach, line 46: outside of the Appium process

PR #860, opened 2023-03-09
feat: Update default component versions

mykola-mokhnach:


KazuCocoa (approved):
I wonder if we should mark as feat (to increase the minor version. not patch) title in the espresso driver since espresso build depends on the dependencies' version...

mykola-mokhnach:
Marked as `feat`

Issue #871, opened 2023-04-19
Can't build espresso server on gradle 8.0


marciogranzotto:
```
[debug] [EspressoDriver@940b (0f9f272c)] Output from Gradle will be logged. To change this, use 'showGradleLog' desired capability
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] FAILURE: Build failed with an exception.
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] * What went wrong:
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] A problem occurred configuring project ':app'.
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] > Could not create an instance of type com.android.build.api.variant.impl.ApplicationVariantImpl.
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]    > Namespace not specified. Please specify a namespace in the module's build.gradle file like so:
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]      android {
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]          namespace 'com.example.namespace'
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]      }
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]      If the package attribute is specified in the source AndroidManifest.xml, it can be migrated automatically to the namespace value in the build.gradle file using the AGP Upgrade Assistant; please refer to https://developer.android.com/studio/build/agp-upgrade-assistant for more information.
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR]
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] * Try:
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] > Run with --stacktrace option to get the stack trace.
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] > Run with --info or --debug option to get more log output.
[EspressoDriver@940b (0f9f272c)] [Gradle] [STDERR] > Run with --scan to get full insights.
[...]

PR #872, opened 2023-04-20
fix: build app with Gradle 8.x

marciogranzotto:
Fixes: #871

On gradle 8.0 and up, the package name needs to be moved from the manifest file to the build.gradle file


mykola-mokhnach:
Thanks @marciogranzotto 

Could you please update the PR title according to https://www.conventionalcommits.org/en/v1.0.0/ ?

marciogranzotto:
@mykola-mokhnach updated the title, I think this is ok, right?

mykola-mokhnach (approved):

Issue #900, opened 2023-08-16
[Espresso] Could not find matching setting. Known setting names are: [driver]


LissaRider:
Hello. I use appium 2 with espresso. I'm trying to connect the ocr plugin, but I can't set the settings for ocr.
Moreover, there are no problems with setting up additional settings for ocr for UiAutomator2.
And so the plug-in is working for both drivers (for Espresso, you still have to change the settings manually in the plug-in index.js file).
"((HasSettings) driver).setSettings(ImmutableMap.of("ocrLanguage", "eng+rus"));" or " desiredCapabilities.setCapability("appium:settings[ocrLanguage]", "eng+rus");" don't work for Espresso

Appium v2.0.1
Available drivers:
espresso@2.25.0 (automationName 'Espresso')
uiautomator2@2.29.3 (automationName 'UiAutomator2')
<selenium.version>4.11.0</selenium.version>
<appium.version>8.5.1</appium.version>

Here's log file:
[appiumServerLog.log](https://github.com/appium/appium-espresso-driver/files/12362492/appiumServerLog.log)

KazuCocoa:
Could you try https://github.com/appium/appium-espresso-driver/pull/901 out?

KazuCocoa:
Could you try out `2.25.2`?

LissaRider:
> Could you try out `2.25.2`?

I've tried it. The same error
[appiumServerLog.log](https://github.com/appium/appium-espresso-driver/files/12370083/appiumServerLog.log)
I have a guess that there might be something wrong with the relationship between the appium itself and the espresso driver, or it might be a java client issue.
I saw a post with the same problem only for another client, but I did not find a similar alternative for Java.
https://discuss.appium.io/t/cannot-access-appium-settings-when-using-espresso-as-an-automation-name/38635

KazuCocoa:
Thanks.
Perhaps https://github.com/appium/appium-espresso-driver/pull/902 is needed. I'll tweak the pr later

KazuCocoa:
What about `2.25.3`?

LissaRider:
Hello. Sorry for the long answer. I was in the hospital.
I tried version 2.25.9 - now everything works. Thank you.

PR #901, opened 2023-08-16
fix: calls this.settings in settings handling

KazuCocoa:
As same as UIA2, maybe should this driver also call `this.settings` stuff to call plugins?
Fixes https://github.com/appium/appium-espresso-driver/issues/900


jlipps (approved):
Ah yes, good idea.

Issue #931, opened 2023-10-23
Version 2.26.4 - Could not find or load main class org.gradle.wrapper.GradleWrapperMain


russtu:
With Appium version 2.2.0 and version 2.26.4 of the espresso driver I get this error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain

PR #932, opened 2023-10-24
Revert "chore(deps-dev): bump semantic-release from 20.1.3 to 22.0.5"

KazuCocoa:
Reverts appium/appium-espresso-driver#916

The same issue reported after https://github.com/appium/appium-espresso-driver/pull/909 occurred in https://github.com/appium/appium-espresso-driver/issues/931

```
kazu@laptop espresso-server % ls
app                 build.gradle.kts    gradle.properties   gradlew             gradlew.bat         lint.xml            settings.gradle.kts
kazu@laptop espresso-server % ./gradlew
Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain
Caused by: java.lang.ClassNotFoundException: org.gradle.wrapper.GradleWrapperMain
```

I haven't dug into the root cause, but it seems like gradlew is broken right now

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "mocha": "^10.0.0",
     "pre-commit": "^1.2.2",
     "rimraf": "^5.0.0",
-    "semantic-release": "^22.0.5",
+    "semantic-release": "^20.0.2",
     "sinon": "^16.0.0",
     "ts-node": "^10.9.1",
     "typescript": "^5.1.6",


KazuCocoa:
```
kazu@laptop espresso-server % ls
app                 build.gradle.kts    gradle              gradle.properties   gradlew             gradlew.bat         lint.xml            settings.gradle.kts
kazu@laptop espresso-server % ./gradlew
Starting a Gradle Daemon (subsequent builds will be faster)

> Starting Daemon
```

after this revert:

mykola-mokhnach:
it is not clear to me how semantic release could affect espresso. Perhaps some files were not included into the release.ignored?
Could you report it as a bug there?

KazuCocoa:
That was `*` in package.json picked up files in the same directories only. So the name's directory was not included. Latest master is ok at least

PR #950, opened 2023-11-08
chore: use omit to prune dev and peer

KazuCocoa:
as same as https://github.com/appium/appium-geckodriver/pull/89


mykola-mokhnach (approved):
@KazuCocoa Feel free to apply the same change to other packages without approval

PR #954, opened 2023-11-23
test: fix npm run test

KazuCocoa:
as same as https://github.com/appium/appium-safari-driver/pull/97

Unit tests work, then this pr is ok.


mykola-mokhnach (approved):
Feel free to marge without approval in other modules where necessary

PR #964, opened 2024-01-07
fix: Import the proper support module

mykola-mokhnach:


KazuCocoa (approved):

jlipps:
Are we sure we want to do this? Previously we had migrated all such imports to `appium/support` so that only one version of `@appium/support` was used: the one inside the Appium peer dep. If we need something that's not yet in `appium/support`, maybe the better strategy is to increase the Appium peer dep version requirement?

mykola-mokhnach:
> to appium/support so that only one version of @appium/support was used

This is exactly what was done. Not sure what your question is about

jlipps:
🤦 i read the diff backwards. sorry, no issue here!

PR #999, opened 2024-05-03
ci: Enable KVM group perms

KazuCocoa:
CI performance is pretty better, yey


.github/workflows/functional-test.yml
```diff
@@ -64,6 +64,14 @@ jobs:
         sudo udevadm trigger --name-match=kvm
     - name: Setup Android SDK
       uses: android-actions/setup-android@v3
+    - name: AVD cache
```
KazuCocoa, line 67: The result did not have any much diff by the AVD caching

PR #998, opened 2024-05-02
ci: use GH Action for detecting LTS Node

eglitise:
Same as https://github.com/appium/appium/pull/20059
Also adjust unit tests to be skipped for Markdown file changes and non-master branches.

PR #1000, opened 2024-05-04
chore: fix type and CI update

KazuCocoa:
https://github.com/appium/appium-espresso-driver/pull/998 + type fix and tweak functional test yml


KazuCocoa:
`Conventional Commits` seems broken now. Maybe recent https://www.npmjs.com/package/conventional-changelog-angular release? it looks like we use existing market place's one. We may need to drop it, or replace with something I guess.
Let's see how it will be for a while..

PR #1007, opened 2024-05-25
chore: set animation scale via settings for api level 26+

KazuCocoa:
https://github.com/appium/appium-adb/pull/745 method instead of `--no-window-animation` for espresso since the flag in espresso instrumentation process did not bring the animation scale back after the process kill.

0/1 for animation scale is to make the behavior as same as `this.settingsApp.setAnimationState`

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
-appium:disableWindowAnimation | To avoid flakiness google [recommends](https://developer.android.com/training/testing/espresso/setup#set-up-environment) to disable the window animation of the Android device under test when running espresso test. Animation state is restored automatically after the session is stopped if it was enabled before it has started. `true` by default
+appium:disableWindowAnimation | To avoid flakiness google [recommends](https://developer.android.com/training/testing/espresso/setup#set-up-environment) to disable the window animation of the Android device under test when running espresso test. Animation state is restored automatically after the session is stopped if it was enabled before it has started. The restore method may not work if the session ends unexpectedly. `true` by default
 appium:timeZone | Overrides the current device's time zone since the driver version 2.38.0. This change is preserved until the next override. The time zone identifier must be a valid name from the list of [available time zone identifiers](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), for example `Europe/Kyiv`
 
 
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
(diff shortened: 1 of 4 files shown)


KazuCocoa:
hm, this option didn't bring back to animation on after the process end.

Below worked on real devices, thus we may need to do this.
```
$ adb shell settings put global window_animation_scale 1
$ adb shell settings put global transition_animation_scale 1
$ adb shell settings put global animator_duration_scale 1
```

README.md
```diff
@@ -188,7 +188,7 @@ appium:chromeOptions | A mapping, that allows to customize chromedriver options.
 Capability Name | Description
 --- | ---
 appium:disableSuppressAccessibilityService | Being set to `true` tells the instrumentation process to not suppress accessibility services during the automated test. This might be useful if your automated test needs these services. `false` by default
-appium:disableWindowAnimation | To avoid flakiness google [recommends](https://developer.android.com/training/testing/espresso/setup#set-up-environment) to disable the window animation of the Android device under test when running espresso test. `true` by default
+appium:disableWindowAnimation | To avoid flakiness google [recommends](https://developer.android.com/training/testing/espresso/setup#set-up-environment) to disable the window animation of the Android device under test when running espresso test. It sets x1 animtion scales after the session ends if the animation was not disabled before the session starts. `true` by default
```
mykola-mokhnach, line 191: Animation is state is restored automatically after the session is stopped if it was enabled before it has started.

lib/driver.ts
```diff
@@ -479,18 +478,29 @@ export class EspressoDriver extends AndroidDriver implements ExternalDriver<
...
   async setWindowAnimationState(isEnabled: boolean): Promise<void> {
     const isAnimationOn = await this.adb.isAnimationOn();
     const shouldDisableAnimation = !isEnabled && isAnimationOn;
     const shouldEnableAnimation = isEnabled && !isAnimationOn;
 
     if (shouldDisableAnimation) {
       this.log.debug('Disabling window animation as "disableWindowAnimation" capability is set to true/fallback to default value "true"');
-      await this.settingsApp.setAnimationState(false);
+      await this.adb.getApiLevel() < 26
+        ? await this.settingsApp.setAnimationState(false)
```
mykola-mokhnach, line 496: would it make sense to move this logic to settingsApp module?

KazuCocoa: Well, i initially thought the same thing but:
- this is only for espresso driver. uia2 driver does it via `--no-window-animation`
     - So uia2 driver won't run `await this.adb.setAnimationScale` to manage the animation stuff only via `--no-window-animation`
- the settingsApp's animation thing should be removed after dropping api level 26, then this disabling/enabling animation thing won't need to be part of the settingsApp
    - Then, this `await this.adb.setAnimationScale` should be only in the espresso driver

So, this is kind of espresso driver specific behavior, thus I haven't added it in the settings module

mykola-mokhnach: I don't think we need to drop the stuff while we support older android APIs. Same has been already done to wifi-related methods: https://github.com/appium/io.appium.settings/blob/a806a2e1e46a45eeb07a0ccf0a9d9e305e3ba734/lib/commands/network.js#L25

KazuCocoa: Okay -> https://github.com/appium/io.appium.settings/pull/176

package.json
```diff
@@ -72,11 +72,11 @@
     "npm-shrinkwrap.json"
   ],
   "dependencies": {
-    "appium-adb": "^12.2.0",
+    "appium-adb": "^12.4.0",
     "appium-android-driver": "^9.6.0",
     "asyncbox": "^3.0.0",
     "bluebird": "^3.5.0",
-    "io.appium.settings": "^5.7.2",
+    "io.appium.settings": "^5.10.0",
```
KazuCocoa, line 79: After https://github.com/appium/io.appium.settings/pull/176

mykola-mokhnach:
actually it would be just a type `chore` change

mykola-mokhnach (approved):

PR #1011, opened 2024-06-13
fix: type error in the build

KazuCocoa:
to fix:
```
lib/driver.ts:689:3 - error TS2742: The inferred type of 'getProxyAvoidList' cannot be named without a reference to 'appium-android-driver/node_modules/@appium/types'. This is likely not portable. A type annotation is necessary.

689   getProxyAvoidList (sessionId) {
      ~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/lib/driver.ts
+++ b/lib/driver.ts
@@ -6,6 +6,7 @@ import type {
   InitialOpts,
   StringRecord,
   SingularSessionData,
+  RouteMatcher
 } from '@appium/types';
 import type { EspressoConstraints } from './constraints';
 import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


lib/driver.ts
```diff
@@ -686,7 +686,7 @@ export class EspressoDriver extends AndroidDriver implements ExternalDriver<
   }
 
   // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
-  getProxyAvoidList (sessionId) {
+  getProxyAvoidList (sessionId): import('@appium/types').RouteMatcher[] {
```
mykola-mokhnach, line 689: this is a .ts file. Simply do `import type`

mykola-mokhnach (approved):

PR #1021, opened 2024-07-10
feat: Publish Espresso server library to JitPack

technoir42:
This is the minimum required configuration to make the library accessible via JitPack.
JitPack sets groupId, artifactId and version automatically. The Maven coordinates of the library will be `com.github.appium:appium-espresso-driver:<tag>`.

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,5 @@
+jdk:
+  - openjdk17
+install:
+  - cd espresso-server
+  - ./gradlew :library:publishToMavenLocal
--- a/espresso-server/library/build.gradle.kts
+++ b/espresso-server/library/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
     id("com.android.library")
     kotlin("android")
+    `maven-publish`
 }
 
 val appiumCompileSdk: String by project
@@ -50,6 +51,22 @@ android {
     testOptions {
         targetSdk = appiumTargetSdk.toInt()
         unitTests.isReturnDefaultValues = true
+    }
+
+    publishing {
+        singleVariant("release") {
+            withSourcesJar()
+        }
+    }
+}
+
+publishing {
+    publications {
+        register<MavenPublication>("release") {
+            afterEvaluate {
+                from(components["release"])
+            }
+        }
     }
 }
 


KazuCocoa (approved):

KazuCocoa:
thank you!

KazuCocoa:
Seems good: https://jitpack.io/com/github/appium/appium-espresso-driver/v3.1.0/build.log

jitpack.yml
```diff
@@ -0,0 +1,5 @@
+jdk:
```
mykola-mokhnach, line 1: I assume we should also add an action to publish the package automatically for each new version (e..g on a new tag event)

KazuCocoa: Meaning... somethig like we create a package to access to the latest package as part of publish, or after publishing npm package?

Jetpack builds a package when it gets an initial request. So to build a package on the jitpack, we need to create a dummy project which runs gradle sync to kick the build on the jitpack project. This is not actually "publishment" like npm, but let jitpack build this repository.

```
	dependencies {
	        implementation 'com.github.appium:appium-espresso-driver:v3.1.0'
	}
```

KazuCocoa: Looks like the GH market has https://github.com/marketplace/actions/autojitpack so probably we could use it

KazuCocoa: I have configured https://docs.jitpack.io/building/#building-ahead-of-time , so not every tag, but every push to master will build snapshot.

mykola-mokhnach:
@technoir42 Would you have some time to document espresso-server-as-library approach?

technoir42:
@mykola-mokhnach I will allocate some time next week to document it

jlipps:
Hi @technoir42 , congrats, the Appium project wants to compensate you for this (and perhaps other) contribution(s) this month! Please reply to this comment mentioning me and sharing your [OpenCollective](https://opencollective.com) account name, so that we can initiate payment! Or let me know if you decline to receive compensation via OpenCollective. Thank you!

technoir42:
@jlipps thank you! my OpenCollective account name is [technoir](https://opencollective.com/technoir).

PR #1022, opened 2024-07-14
feat: Add proper info to the /status API response

mykola-mokhnach:
Now, when the server part could be integrated into the application under test separately, it is important that the driver knows the server version to check it for possible compatibility issues.
Also, the response format itself has been updated to comply with [W3C requirements](https://www.w3.org/TR/webdriver2/#dfn-status)

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,9 +98,11 @@
     "lint:server": "cd espresso-server && ./gradlew lint || cd ..",
     "lint:fix": "npm run lint -- --fix",
     "prepare": "npm run rebuild",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -98,9 +98,11 @@
     "lint:server": "cd espresso-server && ./gradlew lint || cd ..",
     "lint:fix": "npm run lint -- --fix",
     "prepare": "npm run rebuild",
+    "sync-version": "node ./scripts/sync-version.js --package-version=${npm_package_version}",
```
KazuCocoa, line 101: oh, i didn't know of the `npm_package_version`

KazuCocoa (approved):

PR #1038, opened 2025-01-08
chore: bump eslint-config-appium-ts

KazuCocoa:


test/functional/commands/clipboard-e2e-specs.js
```diff
@@ -27,7 +27,7 @@ describe('clipboard', function () {
     const text = await driver.getClipboard('PLAINTEXT');
     try {
       text.should.eql('SGVsbG8=');
-    } catch (AssertionError) {
+    } catch (AssertionError) { // eslint-disable-line @typescript-eslint/no-unused-vars
```
mykola-mokhnach, line 30: I think catching exception by class like here won't work. we need to explicitly call isinstance

PR #1055, opened 2025-03-17
fix: rescue background app

KazuCocoa:
Maybe we accidentally removed this command:

https://github.com/appium/appium-espresso-driver?tab=readme-ov-file#mobile-backgroundapp


KazuCocoa:
-> https://github.com/appium/appium-android-driver/pull/988

PR #1061, opened 2025-05-17
ci: run with Android 16 also

KazuCocoa:
`ContactManager` app had an old target sdk so it doesn't work with Android 14+.
https://github.com/appium/ContactManager/pull/1

All is ok for now.


KazuCocoa:
> App package must target at least SDK version 24, but found 5

Need to bumpd https://github.com/appium/ContactManager for the test case

PR #1083, opened 2025-08-30
test: enable e2e more

KazuCocoa:
Support https://github.com/appium/appium/issues/21426 more...

### DIFF
--- a/ci/run-functional-tests.sh
+++ b/ci/run-functional-tests.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 npx mocha --timeout 10m \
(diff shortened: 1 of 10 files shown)


ci/run-functional-tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+npx mocha --timeout 10m \
```
KazuCocoa, line 3: I'll update to use `*` in the functional-test.yml eventually, but this is for tentative shell script during https://github.com/appium/appium/issues/21426

PR #1094, opened 2025-10-23
feat: Set the minimum supported Android API version to 8/level 26 

mykola-mokhnach:
BREAKING CHANGE: Dropped support of older Android API levels below 26 (Oreo)

### DIFF
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -22,16 +22,10 @@ jobs:
         - apiLevel: 32
           emuTag: google_apis
           arch: x86_64
-        - apiLevel: 30
+        - apiLevel: 29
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -74,14 +74,14 @@
     "npm-shrinkwrap.json"
   ],
   "dependencies": {
-    "appium-adb": "^13.0.0",
-    "appium-android-driver": "^11.0.0",
+    "appium-adb": "^14.0.0",
```
KazuCocoa: https://github.com/appium/appium-espresso-driver?tab=readme-ov-file#requirements needs to bump the min/recommended version

KazuCocoa (approved):

PR #1165, opened 2026-05-22
fix: Kotlin dependencies resolution by dependabot

mykola-mokhnach:
## Summary

Dependabot was only opening Gradle PRs for plugins (e.g. `androidGradlePlugin`) and the wrapper, not for library versions like Espresso (`3.5.1` vs latest `3.7.0`). This completes the version-catalog setup from #1152 by declaring dependencies in `[libraries]` and referencing them via `libs.*` aliases, so Dependabot can map Maven coordinates to catalog versions.

## Problem

After #1152, versions lived in `gradle/libs.versions.toml` under `[versions]`, but `build.gradle.kts` still used string coordinates with runtime variables:

```kotlin
api("androidx.test.espresso:espresso-contrib:$espressoVersion")
```

Dependabot’s Gradle support tracks catalog **`[libraries]`** entries (and `[plugins]`), not bare `[versions]` keys consumed only through Kotlin string interpolation. That’s why AGP bumps worked while Espresso, Gson, NanoHTTPD, etc. did not.

## Solution

- Add a `[libraries]` section to `libs.versions.toml` with full GAV coordinates and `version.ref` links to `[versions]`.
- Switch `:library` and `:app` to `libs.*` catalog aliases instead of string dependencies.
- Add `withCapabilityVersion()` in `CapabilityVersion.kt` so `-Pappium*` overrides from the Espresso driver (`toolsVersions` / `espressoBuildConfig`) still apply at resolve time.
- Update Dependabot config comments to document the expected catalog layout.

No dependency version numbers change in this PR — only how they are declared.

## Test plan

- [x] `./gradlew :library:compileComposeOnReleaseKotlin :app:compileDebugKotlin` (from `espresso-server/`)
- [ ] CI Gradle / unit test workflows pass
- [ ] After merge, confirm Dependabot opens version-update PRs for catalog libraries (e.g. `espresso`, `nanohttpd`, `gson`) on the next scheduled run

## Notes

- `resolveCapabilityVersion()` behavior is unchanged; overrides are applied via `version { require(...) }` on catalog dependencies.
- `composeOnApi` flavor dependencies use `"composeOnApi"(libs.*)` configuration syntax (Kotlin DSL).
- `constraints` block uses inline `version { require(...) }` because the constraint API uses a different receiver than `ExternalModuleDependency`.


KazuCocoa:
hm, code lgtm, but CI keeps failing for passed one in https://github.com/appium/appium-espresso-driver/pull/1164

mykola-mokhnach:
> hm, code lgtm, but CI keeps failing for passed one in #1164

I've checked the logs and the failure is because chromedriver download. Perhaps due to throttling on the server side. It's unrelated to these changes

KazuCocoa (approved):

PR #1175, opened 2026-05-23
feat: Add a script for app diagnosis

mykola-mokhnach:
## Summary

Adds **`diagnose-app`**, a static readiness check for apps that use a prebuilt or embedded Espresso server. It compares the AUT against the driver’s bundled `espresso-server` dependency versions and reports blocking issues, warnings, and suggested `espressoBuildConfig` fragments.

**Usage**

```bash
appium driver run espresso diagnose-app --app /path/to/android-project
appium driver run espresso diagnose-app --app /path/to/debug.apk
npm run diagnose-app -- --app <path>
```

Exit code is non-zero when blocking checks fail.

**What it checks**

- **Gradle project:** manifests (`INTERNET`), `android.useAndroidX`, R8/minify, `InitializationProvider`, `compileSdk` vs Compose requirements, lifecycle-extensions pins, dependency alignment (Compose, Espresso, AndroidX Test, Kotlin, etc.)
- **APK-only:** permissions (via **appium-adb** / `aapt2`), dependency versions from **META-INF `*.version`** (AndroidX/Compose) and **`@kotlin.Metadata` `mv`** (via SDK **dexdump**), obfuscation hints. APK scans can reach **READY** without project sources; Gradle is optional for extra build-file coverage.

**Implementation notes**

- `scripts/lib/diagnose-app-lib.mjs` — orchestration and static checks
- `scripts/lib/dependency-versions.mjs` — version parsing, semver comparison, APK/Gradle scanning
- Uses `appium/support` (`fs`, `zip`, `util.pluralize`, `logger`); no JSON CLI output
- `build-espresso.mjs` — dynamic import of built server module (fixes TS5055)
- README troubleshooting bullet for `diagnose-app`
- Unit tests in `test/unit/diagnose-app-specs.ts`


scripts/lib/dependency-versions/comparison.mjs
```diff
@@ -0,0 +1,171 @@
...
+  if (diff === 'equal' || diff === 'patch') {
+    return {level: 'ok', message: 'Versions are compatible.'};
+  }
+  if (diff === 'minor') {
+    return {
+      level: 'suggestion',
+      message: `Minor version drift. Align the Espresso server via toolsVersions.${mod.toolsVersionKey ?? mod.gradleProperty}.`,
+      espressoBuildConfig: mod.toolsVersionKey
+        ? {toolsVersions: {[mod.toolsVersionKey]: appVersion}}
+        : {gradleProperty: {[mod.gradleProperty]: appVersion}},
```
KazuCocoa: I guess this addresses https://github.com/appium/appium-espresso-driver#espresso-build-config
It has `toolsVersions` so I assume this matches with them, but `gradleProperty` is not. Do you expect users will update gradle files directly instead of the capabilities for the `gradleProperty` case?

scripts/lib/diagnose/project-input.mjs
```diff
@@ -0,0 +1,20 @@
...
+import path from 'node:path';
+
+/**
+ * @param {string} root
+ * @returns {Promise<string[]>}
+ */
+export async function findManifestTexts(root) {
+  /** @type {string[]} */
+  const manifests = [];
+  const candidates = ['app/src/main/AndroidManifest.xml', 'src/main/AndroidManifest.xml'];
```
KazuCocoa: If we specify the candidates here instead of like `fs.glob` for `AndroidManifest.xml`, probably it makes sense to address both paths in https://github.com/appium/appium-espresso-driver/pull/1175/changes#diff-93e4784c9322112a4a7dde0ff8c160d1ed0d3d094bdeef3aa56ea173e778c957R63 (instead of addressing 'module manifests’)

mykola-mokhnach: updated

scripts/lib/dependency-versions/comparison.mjs
mykola-mokhnach: good catch. I've tuned this logic to distinguish between test and prod deps

KazuCocoa (approved):
lgtm

PR #1183, opened 2026-06-21
feat: Migrate package to ESM

mykola-mokhnach:
## Summary

- Migrate the Node package to ESM by enabling `"type": "module"`, adding ESM exports, and switching TypeScript to `NodeNext` module resolution.
- Replace Mocha test execution with Node’s built-in test runner while keeping Chai and Sinon in place.

## Breaking Changes

BREAKING CHANGE: This package now publishes as ESM only. CommonJS `require('appium-espresso-driver')` consumers must migrate to ESM `import`.


eglitise:
Is it not feasible to also drop chai and sinon, like it was done for the drivers already migrated to ESM?

mykola-mokhnach:
> Is it not feasible to also drop chai and sinon, like it was done for the drivers already migrated to ESM?

The test suite is much bigger in this driver. I want to defer that to a different PR

KazuCocoa (approved):

test/functional/commands/mobile-e2e-specs.ts
```diff
@@ -26,14 +25,7 @@ describe('mobile', function () {
   });
 
   describe('mobile:swipe', function () {
-    describe('with direction', function () {
-      before(function () {
-        if (process.env.CI) {
-          // CI env is flaky because of the bad emulator performance
```
eglitise, line 32: I would recommend keeping such comments (here and in the other test files) for future reference, otherwise there is no information on why the tests are skipped

eglitise (approved):

test/functional/commands/mobile-e2e-specs.ts
mykola-mokhnach: I will create a follow up PR later to verify and reenable some e2e tests that are stable enough to be executed in CI

PR #1208, opened 2026-07-29
chore: Drop chai

mykola-mokhnach:


test/functional/commands/touch-e2e-specs.ts
```diff
@@ -282,8 +279,8 @@ describe('touch actions -', {skip: SKIP_TOUCH_TESTS, timeout: E2E_TEST_TIMEOUT},
       beforeEach(async function () {
         await startTextSwitcherActivity();
 
-        expect(await driver.$("//*[@text='0']")).to.exist;
-        await expect(driver.$("//*[@text='1']")).to.be.rejectedWith(/NoSuchElement/);
+        assert.ok(await driver.$("//*[@text='0']"));
+        await assert.rejects(async () => driver.$("//*[@text='1']"), /NoSuchElement/);
```
eglitise, line 283: Is the extra `async () =>` needed?

test/functional/commands/touch-e2e-specs.ts
```diff
@@ -311,11 +308,11 @@ describe('touch actions -', {skip: SKIP_TOUCH_TESTS, timeout: E2E_TEST_TIMEOUT},
           },
         });
 
-        expect(await driver.$("//*[@text='1']")).to.exist;
+        assert.ok(await driver.$("//*[@text='1']"));
       });
 
       it('should do touch/doubleclick event', async function () {
-        await expect(driver.$("//*[@text='2']")).to.be.rejectedWith(/NoSuchElement/);
+        await assert.rejects(async () => driver.$("//*[@text='2']"), /NoSuchElement/);
```
eglitise, line 315: Same here

eglitise (approved):

test/functional/commands/touch-e2e-specs.ts
mykola-mokhnach: Yes — `driver.$()` returns webdriverio's `ChainablePromiseElement`, which isn't structurally a `Promise`. `assert.rejects` is typed as `(block: (() => Promise<unknown>) | Promise<unknown>, ...)`, so passing the chainable directly (or wrapping it in a plain non-async arrow) fails to compile with TS2769. The `async () =>` wrapper coerces the return type into a real `Promise`, which is what makes it type-check. Verified locally: removing `async` breaks `tsc -b`.

test/functional/commands/touch-e2e-specs.ts
mykola-mokhnach: Same reason as above — `driver.$()` returns `ChainablePromiseElement`, not a `Promise`, and `assert.rejects`'s type signature requires `(() => Promise<unknown>) | Promise<unknown>`. Dropping `async` here fails to compile with TS2769 too.

PR #1213, opened 2026-08-22
feat: add missing support for otherApps capability

eglitise:
Similarly to #1212, the README lists the `appium:otherApps` capability, but it is only implemented in UiAutomator2. This adds similar support based on the UiA2 implementation here: https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/aut.ts#L44

Note: UiA2 also supports apps with the `.apks` extension. I've not added it here, since the driver does not handle this extension anywhere, but I can add it if such apps should also be supported.


KazuCocoa (approved):

mykola-mokhnach:
Theoretically this cap is supported, but practically it's mostly useless as Espresso cannot automate anything beside application sandbox, unlike UIA2 driver.

lib/commands/app-install.ts
```diff
@@ -134,6 +134,18 @@ export async function initAUT(this: EspressoDriver): Promise<void> {
...
 
+  // Install any "otherApps" that were specified in caps
+  if (this.opts.otherApps) {
+    let otherApps;
+    try {
+      otherApps = utils.parseArray(this.opts.otherApps);
+    } catch (e) {
+      throw this.log.errorWithException(`Could not parse "otherApps" capability: ${(e as Error).message}`);
+    }
+    otherApps = await Promise.all(otherApps.map((app) => this.helpers.configureApp(app, [APK_EXT])));
```
mykola-mokhnach: consider assigning the result to a new variable

eglitise:
> Theoretically this cap is supported, but practically it's mostly useless as Espresso cannot automate anything beside application sandbox, unlike UIA2 driver.

I see this capability mostly as a way to enhance testing shared data between multiple apps from the same vendor. I've tested such in the past, although not with this driver.

lib/commands/app-install.ts
eglitise: Done

KazuCocoa:
Yes, it's a valid scenario. I did long before with vanilla espresso

mykola-mokhnach (approved):

PR #1220, opened 2026-09-01
fix: do not build espresso server on npm install

mykola-mokhnach:
The prepare script ran a full rebuild (node + Gradle server), making package installs unnecessarily slow. Also fixes clean:node, which was accidentally invoking the full build script and re-pulling in the server build through rebuild:node.


eglitise (approved):
Thanks. I also ran into this issue while making the docs, but I wasn't sure what to safely change so that CI and publishing don't break.

PR #1221, opened 2026-09-05
refactor: stop relying on the deprecated BaseDriver.getSession

mykola-mokhnach:
Appium's base-driver is dropping getSession entirely, including from the ExternalDriver interface. AndroidDriver's inherited getSession is bound to AndroidDriverConstraints, which was never assignable to EspressoConstraints, so this override existed purely to satisfy the typechecker. Return this.caps directly instead of delegating to the soon-to-be-removed BaseDriver.prototype.getSession.


eglitise:
Won't this cause the method to omit event timing information?

mykola-mokhnach:
> Won't this cause the method to omit event timing information?

please don't check this PR yet, it's in progress

mykola-mokhnach:
> > Won't this cause the method to omit event timing information?
> 
> please don't check this PR yet, it's in progress

It's ready now

lib/driver.ts
```diff
@@ -226,10 +228,6 @@ export class EspressoDriver extends AndroidDriver implements ExternalDriver<Espr
     return appInstallCmds.isAppOnDevice(this);
   }
 
-  override async getSession(): Promise<SingularSessionData<EspressoConstraints>> {
-    return await BaseDriver.prototype.getSession.call(this);
-  }
-
   // needed to make the typechecker happy
   async getAppiumSessionCapabilities(): Promise<SessionCapabilities<EspressoConstraints>> {
```
eglitise: Maybe this can be removed too, similarly to the UiA2 driver PR?

mykola-mokhnach: Good catch, removed — `AndroidDriver<EspressoConstraints>` already types `getAppiumSessionCapabilities` correctly via `BaseDriver`, so the pass-through cast wasn't needed.

eglitise (approved):

PR #1227, opened 2026-09-21
fix: use correct locatorStrategies

eglitise:
I was checking the defined value of `this.locatorStrategies` across various drivers, and it seems the Espresso driver is the only one that doesn't define its full list. This fixes it, and makes the list correctly depend on the subdriver as well.

Since the find element endpoints are proxied directly to the Espresso server, these changes don't actually affect anything yet. Still, I was able to also integrate them by implementing `doFindElementOrEls` similarly to UiAutomator2, and they worked as expected. The reason why this implementation is not included here is because it changes the error response when using an unsupported strategy, which would be a breaking change.


lib/driver.ts
```diff
@@ -439,9 +459,15 @@ export class EspressoDriver
...
   }
 
-  async onSettingsUpdate() {
-    // intentionally do nothing here, since commands.updateSettings proxies
-    // settings to the espresso server already
+  async onSettingsUpdate(key: string, value: any) {
+    if (key === 'driver') {
+      const driverValueLowerCase = String(value).toLowerCase();
+      if (driverValueLowerCase === 'espresso') {
+        this.locatorStrategies = LOCATOR_STRATEGIES_ESPRESSO_CONTEXT;
```
mykola-mokhnach: should this be just universal fallback (e.g. `else this.locatorStrategies = LOCATOR_STRATEGIES_ESPRESSO_CONTEXT;`) ?

eglitise: updated

mykola-mokhnach (approved):

